PR #3, opened 2018-04-13
support for zipped log parsing

amgine:
.NET has native support for zip archives, and evtc logs are usually compressed, so parsing compressed logs is ez


baaron4:
works out ok ty

PR #13, opened 2018-05-18
More visibility stuff

EliphasNUIT:
I've rewritten createxxxTables in the same way as createHTML. Also fixed a few more html errors (present and possible).

Missing createPlayerTab and createBossSummary, I'll do it this WE.

### DIFF
--- a/LuckParser/Controllers/Controller1.cs
+++ b/LuckParser/Controllers/Controller1.cs
@@ -3948,7 +3948,7 @@ namespace LuckParser.Controllers
             }
             if (presMech.Count() > 0)
             {
-                sw.Write("<script> $(function () { $('#mech_table').DataTable({ \"order\": [[2, \"desc\"]]});});</script>");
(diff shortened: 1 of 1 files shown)


baaron4:
Everything is checking out ok, again thanks for the cleanup. Id guess your doing it in part to just catchup to how the code works.

EliphasNUIT:
Yeah exactly. It aslo helps me find small bugs here and there. As soon as I'm done (aka today) I'll start looking at Deimos and this week I've also found some issues with CM bosses (like for example Deimos CM failure registering as a success with a fight of 0 secs).

baaron4:
Oh about Deimos. Hes got a special check for if he was killed or not. Since the last 10 percent he changes form I have to find the instid of that final form and see if that was killed. But since the final form is a structure Im not sure if Im fully capable of finding it. So at the start of Deimos logs Im just saying if(deimos)(set result as N/A)

PR #16, opened 2018-05-23
Several stuff

EliphasNUIT:
- Some bug fixes here and there
- Put down baseground for foods/utilities
- Changed the progress bar on the header a little. Full health displayed below, damage done in green progress bar and failed in red bar (multi progress bar). Damage done and remaining health on hover
- Huge type rework in order to manage overflows (changed to ulong, long, short or ushort where applicable). Seems to have fixed overflow issues (at least on my test database) but could use more testing
- Nonshareable buffs are tracked per player when the 'all buff' option is enabled. They are displayed on the player tab, on the graph. Don't know if they should be put on a table, lots of them are more interesting to see on a time bar than plain uptime %
- Fixed a bug on Deimos CM, where the structure boss seems to appear long before the last 10% (before disapearing again, seems to happen at the start of the fight) thus if the pull is a fail, lastAware of the boss becomes nearly identical to firstAware, causing dps explosion and inevitable overflows
- Extended StateChange with more ids and migrated the successfull kill checking to "bouncy chest" method. Removed getBossKill() in Controller.cs as a result. Tested on several bosses, including SH. Seems to be working fine

### DIFF
--- a/LuckParser/Controllers/Controller1.cs
+++ b/LuckParser/Controllers/Controller1.cs
@@ -67,7 +67,7 @@ namespace LuckParser.Controllers
 
             return byt;
         }
-        private int getShort()
+        private ushort getShort()
         {
             byte[] bytes = new byte[2];
             for (int b = 0; b < bytes.Length; b++)
@@ -216,7 +216,7 @@ namespace LuckParser.Controllers
             safeSkip(1);
 
             // 2 bytes: boss instance ID
-            int instid = getShort();
+            ushort instid = getShort();
 
             // 1 byte: position
             safeSkip(1);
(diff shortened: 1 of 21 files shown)


baaron4:
Non shareable buffs have always been relegated to just the graph since its mostly just things you would like to know if where present in a rotation.

Deimos is quite ugly glad to see you found a pretty good fix.

As for bouncy chests I wasnt sure if it would always be a good way to check, because you may still have unopened chests from previous kills/maybe chest overflow/maybe its your second time killing the boss that week. However If it works Ill stick with it. Since it solves the issue of some bosses not actually dieing at the end or changieng ids 

good work

EliphasNUIT:
Well you always get the bouncy chest after a kill (not the big chest) si it's always there at the end. As far as I know, overflow or a bouncy chest beeing there before, the event should still register. 
For now I've tested this method on several logs, by me or by friends (30 ish logs) and it seems to be working fine. I'll try to keep the bouncing chests unopened this week and check if it still works.

PR #22, opened 2018-06-03
Bugfix/boons

EliphasNUIT:
A huge boon rework, it should fix every bug we had with them so far:
- Moved some files around, purely for cosmetics reasons, ParseModels was getting too crowded
- Remove Statistics, Utility and BoonFactory, their use was split between boon.cs and player.cs
- Renamed AbstractBoon/Duration/Intensity to BoonSimulator/BoonSimulatorDuration/BoonSimulatorIntensity. Using you stack management and the new BoonSimulationItems, I simulate the whole fight from the point of view of the boon. The result is then stocked in Player.cs as a BoonDistribution. This is done only once per player and from it you can get boon uptimes, boon gen for self, group, offgroup and squad and overstacks, also speeding up the html generation process.
- The new method is also "time interval" friendly, we can trivially extend it to phases once we get there.

I'll heavily document this process when I've the time. I still thought  I should push this now, as the fixes are pretty important and release worthy.

Bug fixes:
- Fixed a bug where hours would always appear in duration
- Fixed a multi threading bug
- Fixed some incorrect boon data

Found issues:
- I've found a bug where damage dist tables for minions were incorrect, investigating


baaron4:
Im gonna do a bit of testing and then probably push the project to v1.0 release if everything works out ok. Thank you for the good work! I knew boons needed a rework just havent had the time to do a full relook at it

baaron4:
Still looks like we are 5 to 7% lower boon uptime then dps.reports. Im starting to wonder if maybe RaidHeroes is just wrong?

EliphasNUIT:
I've also noticed this but I believe those are raid heroes issues
https://dps.report/ObZc-20180530-205340_mo here you can see me, Huyana, having 99% fury uptime when I died at 73% of the fight, whereas GW2EI shows 71,6%

baaron4:
His uptimes might be out of player uptime rather than boss uptime. However the issue Im encountering is that for things like pinpoint distribution I can see that the buff just drops off the engie twice for about 3 seconds 
[20180529-213713.evtc.zip](https://github.com/baaron4/GW2-Elite-Insights-Parser/files/2066313/20180529-213713.evtc.zip)

https://dps.report/Y8Iz-20180529-213713_vg

I removed EIs ability to remove boons on buff remove and it kind of fixed it (other boons were still 7% percent off). I think this has to do with the way buffs are reapplied. Ive noticed when using the addon that the buff applied then removed in the combat log. However we are currently removing all instances at the moment a buff remove event happens deleting that application essentially.

EliphasNUIT:
If it goes by player uptime then it could explain the differences. There are some stranges occurences I've seen for example Razorclaw's Rage (renegade buff): in a log apparently the renegade's allies remove this buff from himself (like renegade applies 1 sec of buff to himself, the warrior in his group removes 0.7 secs of buff from the renegade). I'll be investigating

EliphasNUIT:
I can confirm that removing the MANUEL part from getBoonMap gives moer logical results. The boon graphs between raid heroes and GW2EI seems to be the same.
Strangely enough, if I put 9 max stacks for Quickness and Retal (old max stack values) then it's nearly a match. A this point I don't know what to say without seeing their code but the MANUEL part in our BoonMap is definitively problematic

baaron4:
problem is: 
* Ive seen buffs that execute manual buff remove then iterate 23 singles removes 
* profession buffs like Sadistic Searing only proc Manuel remove when casting a skill
Arc says to disregard Manuel when possible but its hard to tell when its the right buff to

PR #25, opened 2018-06-05
Bugfix/boons

EliphasNUIT:
Some new stuff:
- I have a solution for MANUAL thingy we have discussed, the results are much better but may need more testing to be sure that the table is correct.
- I did a few test yesterday and we are around 1-2% below raid heroes, at this point I believe we are looking at rounding errors
- Fixed a bug where the duration of the fight was higher than expected
- Some qol changes so that the boon order on the graph is the same as before (and not ordered by source)

### DIFF
--- a/LuckParser/Controllers/Controller1.cs
+++ b/LuckParser/Controllers/Controller1.cs
@@ -506,17 +506,6 @@ namespace LuckParser.Controllers
                     log_data.setLogEnd(c.getValue());
 
                 }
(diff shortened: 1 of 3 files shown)


baaron4:
This looks good Ill try and find time to work on and test the differences. 

I started a new job recently and have very little time until I get off (8 days on 6 days off) so I may be a bit slow in responding for a while

Issue #26, opened 2018-06-07
Average Boon amount


DerpyMoa:
Hello,
Would be really useful for classes with boon mods to be able to see average boon amount in player summary or boon tab.

EliphasNUIT:
This should be easily doable. Would it be okay for it to appear as a tooltip while hovering over the character name on the boon table? I can also put it on the boon graph of player tab to see its real time evolution if you think it's interesting.

EliphasNUIT:
There you go:
![avgboons](https://user-images.githubusercontent.com/16018765/41097009-d0cadafc-6a56-11e8-90b0-b02712977e02.png)
And
![avgboons 2](https://user-images.githubusercontent.com/16018765/41097288-bb89730a-6a57-11e8-973a-fae5e2171a12.png)

DerpyMoa:
Looks perfect. This way can check amount of boons during Weaver burst too.

PR #27, opened 2018-06-08
Maintenance

EliphasNUIT:
Feature:
- Avg number of boons when hovering the name of the player on boon uptime table
- Number of boons on a given time on the player tab (as Number of Boons)

Bugfix:
- Fixed a bug where Aegis and Stability were not properly removed after a hit/cc
- Fixed a bug where the filtering in get boon map was too agressive
- Fixed a bug where the overstack values were too high and associated to incorrect players

You can also close Derpy Moa's issue as Enhancement


baaron4:
#26 Issue related

PR #35, opened 2018-06-14
Feature/phase

EliphasNUIT:
Mostly phase detection stuff, I'm still missing KC but it should be ok by tomorrow. The code is good to go and I'd like people other than me to start testing^^

A lot of refactoring and made a lot of stuff more generic, code size reduced by nearly 15%:
- Created ParserHelper and HTMLHelper and put some stuff from Controller on it
- Player and Boss reorganized, they are under AbstractMasterPlayer (entities that can have minions)
- Minions are an abstract player now and have casting logs
- Damage dist tables normalized between players/boss and minions
- Added boss health graph to boss summary table
- Added friend/foe check to buff removals (Boons can only be removed foes and condition can only be cleansed by friends, manual is also friend only now)
- Removed unused usings

Bugfixes
- Fixed a bug where gadgets would be put in NPC tables (an agent parsing bug). This was causing total dps explosions on some bosses (mainly SH and VG)
- Fixed a bug where the last point of the boon graph would'nt be drawn

Known issue:
- There is still some noise in total DPS (small but noticeable when checking damage dist table). Mainly for minion based classes (Ranger and Mesmer) their minions cast sometimes stuff the shouldn't be able to (like spirits doing mesmer auto attacks). Investigating.


baaron4:
looks good. Lets see if we can find any bugs

PR #37, opened 2018-06-15
Background worker upgrade

QuiCM:
Hiya

This PR makes some fairly significant changes to the core UI:
* Concurrent file parsing
  * Each log gets its own background worker
  * Each background worker runs independently of the others
  * Parse multiple logs at a time!
* Replaced the ListView with a DataGridView. This lets you embed a button inside a column (see images below)
  * DataGridView has a data binding to a collection of `GridRow` objects. Update the data and the view updates too :)
* Renamed "Parse" to "Parse All" and "Cancel" to "Cancel All"
  * These affect every log in the view
* Individual logs can be removed by click on the grid row and pressing 'delete'

Empty view:
guildwars2eliteinsights_2018-06-15_15-08-32

View with two logs, no parse actions taken
guildwars2eliteinsights_2018-06-15_15-09-51

View with two logs, one parsing & one cancelling
guildwars2eliteinsights_2018-06-15_15-10-45

View with two logs, both parsing
guildwars2eliteinsights_2018-06-15_15-11-22

View with two logs, both completed
guildwars2eliteinsights_2018-06-15_15-14-36


QuiCM:
Note: 
I've only tested with the two logs I have available currently. Save to CSV not tested. Different settings not tested.
However, this change only touches UI elements and shouldn't break anything related to the actual parsing

EliphasNUIT:
this looks cool.

@baaron4 Could you do a little release before accepting this PR so that people can get phase detection fixes and KC. I'd also like to test this a little bit before making it accessible to the users.

baaron4:
had enough time to make a setting for toggling the phase detection and making a release. Ill merge this and release it once it gets the conflicts fixed up

EliphasNUIT:
Console generation has been removed though

PR #38, opened 2018-06-16
some refactors

cordbleibaum:
Hi,

I'm preparing some of the code to rework report generation a bit. I've done a couple refactors:

- Seperated HTML-building and CSV-building into classes. I've mostly took the code and moved it
- Use container for settings instead of array to improve readability

I'm planning to rework the generation of html and csv and possibly other formats(json?) by first generating statistics and building a report afterwards. Also I'm thinking about adding custom theming for the reports. A further project could maybe be to use js to render the report and avoid writing as much html in c# making it easier to use themes and templates.


EliphasNUIT:
Nice, I'll take a deeper look at this when I get home.

baaron4:
Mostly just refactoring and renaming shouldn't be an issue so merged. I look forward to seeing any future changes!

PR #40, opened 2018-06-17
Major bug fix

EliphasNUIT:
I have changed the way agents were parsed to do it the way Delta documented.

Added the concept of master_agent. The issue we had with total dps and minions in general was related to the fact that minions could share the same ids for example: you create a clone (id 508), the clone dies, later the ranger spawns a spirit and it can have the same id. That would create bad association between boss, adds, minions and player, provoking dps explosion on some instances.

For the minions, we check the agentID and not the instanceID. And for players, bosses and minions, we always check that the treated combat item is in the range the actor was aware.

The fix is important, i'll do some cleanup later.
Edit: done, also corrected a reg in boons, I should have corrected the MANUALFRIEND back to MANUAL after the dst/src inversion fix for buff removal


baaron4:
Not seeing any issues just improvements !

PR #41, opened 2018-06-18
Background worker upgrade 2.0

QuiCM:
* Cancellation is now more responsive
* Console parsing split into a separate class


baaron4:
- [ ] "Open" Button after log finishes parsing not working in any condition
- [ ] Canceling throws exceptions occasionally, 
- [ ] Will  probably make the slower (1 by 1) option default since allowing the program to use as many background workers can easily lead to a user using 100% of their cpu
- [ ] CSV wont write if both Outputs for HTML and CSV are checked

QuiCM:
1) Oops, will fix that in a new commit
2) Do you know which exception?
3) fair enough
4) This is the current functionality too

baaron4:
I'll try and find some exceptions after work tomorrow, but I got most by just cancelling the log in parsing phase. 

Logs used to parse both if both were check marked. Saved time for users looking to get all files generated. I think we lost that functionality with a merge at some point

PR #42, opened 2018-06-18
Feature/10s30s

EliphasNUIT:
As the name indicates, 10s and 30s graphs. Also added settings for it.


baaron4:
Only Issue I see is that mechanics still line up with total dps on alternate 10s/30s graphs. but really it doesnt matter. Could prob just remove them honestly.
Merging since its good to go

PR #43, opened 2018-06-18
Fractals

EliphasNUIT:
First drop for fractal support. Some mechanics id and phase detection for all CM bosses.

Requires testing though, I don't have a CM fractal log under my hand right now.


baaron4:
[fracLogs.zip](https://github.com/baaron4/GW2-Elite-Insights-Parser/files/2113569/fracLogs.zip)
Here the few that I have have. Only issue I saw was some datatables didnt form up. More than likely just mechanic tables not being built properly

PR #45, opened 2018-06-19
Fix: strip extension

cordbleibaum:
Strips extensions from filename again, assuming the change was not intended.


EliphasNUIT:
Thanks! I'll accepting this, only two lines have changed, no need to wait for @baaron4

PR #47, opened 2018-06-19
Seperate html and generation of statistics

cordbleibaum:
I've separated most of the calculation of statistics and html. Currently separated:
- Final DPS
- Boons
- Misc Stats
- Defenses
- Support
- Conditions
- Graphs, moved to GraphHelper, as it is neither final phase/boss statistics nor html specific

Also I've added switches for the statistics calculator to avoid generating too much unnecessary data.


EliphasNUIT:
I've tested this and haven't found any regression, thanks.
I'll merge.

PR #49, opened 2018-06-20
Maintenance

EliphasNUIT:
Improvements:
- Added post patch last phase Deimos id alongside the old one.
- Better agent filtering.

Bugfix:
- Deimos, strangely, has multiple agents during the fight(like 3-4), at least on the log Derpy sent me. This causes a total dps explosion as the boss dps is added multiple times as the boss id exists multiple times in our agent list. I did a generic method which cleans the NPC list for the bossid should we detect multiple instance of the boss.


EliphasNUIT:
Merging.

Issue #51, opened 2018-06-20
Console output is not written to stdout


micksam7:
This has been a big pain and I've written a lot of hacks to get around it, but they all broke again.

Elite Insights doesn't write to a console's stdout. It behaves in one of two ways:
1) Spawns a new thread, releases control to the console, and then hooks the background console window from the new thread, breaking stdout pipes.
2) Spawns an entirely new console window and writes there where nothing can get to it.

Previously I've been able to spawn a pipeable console instance, launch Elite Insights inside of it, and read from it there, but #36 broke that. This approach now simply causes Elite Insights to spawn a new console window.

I realize this is primarily a C# and winforms issue, but some sort of resolution would be appreciated!

EliphasNUIT:
I'll rollback on that. I was trying some stuff as getting proper console output is a mess on windows application.

EliphasNUIT:
Fixed with #53

PR #52, opened 2018-06-21
Maintenance

EliphasNUIT:
Fixed regression on #51


cordbleibaum:
Found a way to fix this, without removing AllocConsole
#53

Issue #50, opened 2018-06-20
Configuration through console


micksam7:
Would it be possible to set configuration options through the console for batch use? Even if it's simply specifying a different .config file to use?

EliphasNUIT:
Done with #54

PR #54, opened 2018-06-21
Configfile switch

cordbleibaum:
Added a switch to set a different configfile
#50


EliphasNUIT:
This looks great. Merging.
Thanks

PR #61, opened 2018-06-24
Light theme

cordbleibaum:
Added a light-theme option and used another bootswatch theme as light theme


baaron4:
May consider swapping if(light) logic to a switch(theme) one. Then instead of theme toggle make a theme drop-down.allows for growth.probably not necessary but hey why not lol. Will also merge when I can test.

PR #62, opened 2018-06-25
performance improvements

amgine:
refactored quite a bit to improve parsing speed; expected overall performance gain on single log should be around 35%-45%


amgine:
no, this patch does not change any logic, it just improves parsing speed

baaron4:
Scrolled through all the change on mobile and it looks solid. Once I get home I'll give it a test parse then merge.

EliphasNUIT:
I'll take a look at the golem tomorrow, though the problem comes from the logs, not the parser (that's why it is a hack ^^)

PR #64, opened 2018-06-26
Maintenance

EliphasNUIT:
I was also working on movements ^^
So:
- moved movement data from player to AbstractMasterPlayer so that we can have the info on bosses too
- moved its generation from fill mising data to AbstractMasterPlayer.setMovements()
- added movement_data to parsed log
- the generation is controlled by statistics calculator, so that we can easily control it with settings
- added time data to Point3D

I've also moved boon related stuff to AbstractMasterPlayer


EliphasNUIT:
Merging this, just some refactoring. I've disabled movements parsing for now to not negatively affect performance for now

PR #84, opened 2018-07-19
Combat replay

EliphasNUIT:
Improvements:
- Removed some unecessary parameters (Player graph totals and player graph boss, as they are basically the same thing as the ones for the dps plots)
- Made dps graph a little bit bigger so that the legend looks less crowded
- Removed self from group and squad gen tables so that you can see what a player contributed to its group/squad without numbers beeing skewed by self generation
- Removed some hacks and implemented proper solutions
- Little performance improvements
- We can manage Dhuum logs that come without the pre-event now
- Typo fixes in mechanic naming
- Added contributors to readme
- Major improvement: added EffectStacking logic classes to boons, so that we can easily change the update logic of the buffs. Currently supported: QueueLogic (Boons and Conditions), OverrideLogic (Buffs) and HealingLogic (Regeneration)

Bug fixes:
- Fixed a bug where sorting profession columns wouldn't work
- Fixed a bug where fights would start later than expected

Features:
- Added cleave only dps plots (under option, false by default)
- Added boon table to boss tab when applicable: if the boss receives a boon different than retal, basically relevant in fractals, Dhuum and Xera
- Added split phases for VG, Gorse, Sab and Samarog

Major Feature: Combat Replay
- Under options (disabled by default)
- Nice looking arena maps (massive thanks to Linus)
- All fractal CM bosses and raid bosses supported
- Basic time control and player selection (other players in the selected's group are highlighted in blue)
NB: some arenas may be slightly off, getting it to pixel perfect level is no easy task. If you find something severely off please contact us. We want to do more stuff on it (like range circles, current boon/dps situtation, rotation etc) but I believe the current status to be good enough for a public release (and I'd like to stress test it so that we can find and fix bugs on it). Feature requests on it are obviously welcome.


baaron4:
Went over everything in VS and it looks good. Going to work on some CSV thing today then release later

PR #101, opened 2018-08-08
Some improvements

EliphasNUIT:
Improvements:
- added phases to statistics
- improved some mechanics sizes on combat replay
- combat replay supports doughnut type actors now, still missing link type
- added Xera's Fury, Zealous Benediction, Outer Vortex Slash, Gorseval's Rampage (the size of that can be wrong, still looking into it) to combat replay. 

@baaron4 regarding you stacking distance dev, I've made some changes to combat replay to better accomodate your intentions and reduce the number of tests you have to do:
- Players and bosses will always have the same quantity of positions by forcing the interpolation (adds and other stuff can still have only one position when tracked)
- Combat replay was already tracking downs and deads (in setAdditionalData in Player.cs), added dc to that list
- Implemented a getActivePositions() to Combat replay that return the list of positions with nulls when the player was dead or dc

Is that okay with you?


baaron4:
Ya that looks ok. I'll find the time to use those methods to improve distance stats performance

PR #111, opened 2018-08-18
Tweaks on Foods, Markers, IDs and Mechanics

LunusLupus:
Merge issue from the change to BossLogic -> FractalLogic, should be fixed though.


EliphasNUIT:
Looks good, though what you have added in Boon.cs seem to have a huge indentation, could you fix it to be aligned with the rest please? :P
You can add instability related stuff in FractalLogic if you wish (each fractal bosses will call it).

LunusLupus:
That's interesting, indentation is perfectly normal on my local branch. Looks like different space conventions on tab. 8 spaces per tab is quite unusual. ;)
Will change it to spaces then.

LunusLupus:
From the looks of it, that's the ONLY file that gets displayed with 8space indentation tabs for some reason. Any idea why?

PR #113, opened 2018-08-22
C# conventions

EliphasNUIT:
C# convention:
- 98% (maybe 99%, did not really check) of renaming stuff to respect C# naming conventions
- Removed unnecessary qualifiers
- Migrated delegate functions to lambda

I'll do a pass on our get/set methods to migrate onto properties when applicable


EliphasNUIT:
Let's make sure the conventions are respected from now on ^^

PR #115, opened 2018-08-24
Small refactorings

Flomix:
Changed the SkillInfo storage from List to Dictionary, because it is mainly used to find a skill by id. Iterating the list every time sucks.

Added constants for often used IDs like dodge, resurrect, weapon swap etc.


EliphasNUIT:
Looks good!

PR #128, opened 2018-08-27
Initial JSON Report Support

cordbleibaum:
Added JSON output:
 - option in settings
 - statistical information from statistics class
 - mechanics
(a bit more info than in csv)

Things that will be built in later commits:
 - data for graphs in json
 - bson

Things that could be done:
 - serializing the JsonLog data structure to xml and renaming it to SerializableLog
 - json could be used for other apps and databases as input


EliphasNUIT:
I've been able to play with this a little. Looks good. 

However what's the reason behind puting stackCenterPositions (they are just used to get StackDist) ? It makes the generated jsons huge and we don't even expose it in html and csv.

Another thing, for buff uptimes, it puts the id of the boon instead of the name so you have "740": instead of "Might": (could go a Boon.GetAll().GroupBy(x => x.GetID()).ToDictionary(x => x.Key, x => x.Select(y => y.GetName()).First()))

cordbleibaum:
stackcenter: I thought I've removed it, but apparently that change got missing somewhere

I'm changing the buffs to use the name, is probably better

kcasctivcrmn:
YEAH BROTHER!!!

PR #135, opened 2018-08-31
Add Travis CI

Sejsel:
This adds a very simple CI that simply checks if the solution builds. No tests are being ran (we don't have any, but at the very least adding a few logs to test if they can be parsed would be nice for the future).

Merging this is not enough, somebody with owner access to the repository also needs to log in at https://travis-ci.org/ and add the repository, then it will be ran for commits in branches and pull requests.

Travis CI is free for open source projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: csharp
+solution: LuckParser.sln


EliphasNUIT:
Regarding logs we'd need to ask the permission of all the participants first before pushing it to the repo, but yeah would be nice.

EliphasNUIT:
@baaron4 as the owner, you are the only who can add to travis ^^

PR #138, opened 2018-09-02
Mechanics

LunusLupus:
* Adjusted some mechanics on MO and SH 
* Corrected every mushroom in the Sloth map
* Adjusted Breakbar mechanics for all bosses to have more intuitive/comparable markers and added successful breaks (tricky in case of Samarog and Matthias, kills may also show that the CC ended)
* Changed the SpecialConditon to allow checks of different attributes than 'value'
* The special condition check is now a delegate that can be used in the form of x => x.getCombatItem().<CombatItemAttribute>, where CombatItemAttribute can be 'Value' for example
* For 'EnemyHit' and 'SkillonPlayer' type mechanics getCombatItem() has to be replaced with getDLog()
* Changed existing mechanics to the new format
* Added 'Downed', 'KillingBlow' and 'Interrupt' as eligible for the 'IsHit()' check
* Updated Boon.cs with the revenant changes


LunusLupus:
Theoretically, yes. But failure to CC her is a guaranteed wipe and the end of the fight (there is no way to circumvent this), this is rather to see how fast she was broken. But I can remove it completely or change it if you still feel we should keep it?

EliphasNUIT:
True, but putting the fail as a mechanic would make it clear at first glance that the wipe came from it whereas on the other end you'd have to check if both are present (start and successful end) on the plot, or the numbers for both are the same on the table

LunusLupus:
I could also just keep both? Make it red if failed, similar to how it's handled on the other bosses, but also make a successful check if it's broken? Actually, I could add that to all bosses probably so people can check how good/fast their CC was. What do you think?

EliphasNUIT:
That would be great yes!

LunusLupus:
Ok, will implement those tonight/tomorrow.

LunusLupus:
Yes, I can change it. That's just the name the game gave to the buff. But it's from the Phase Traversal skill.

LuckParser/Models/BossLogic/Arkk.cs
```diff
@@ -23,23 +23,25 @@ public Arkk()
...
+            new Mechanic(791, "Fear", Mechanic.MechType.PlayerBoon, ParseEnum.BossIDS.Arkk, "symbol:'square-open',color:'rgb(255,0,0)',", "Eye","Hit by the Overhead Eye Fear", "Eye (Fear)",0,(condition => condition.getCombatItem().Value == 3000)), // //not triggered under stab, still get blinded/damaged, seperate tracking desired?
+            new Mechanic(39645, "Breakbar Start", Mechanic.MechType.EnemyCastStart, ParseEnum.BossIDS.Arkk, "symbol:'diamond-tall',color:'rgb(0,160,150)',", "Breakbar","Start Breakbar", "CC",0),
+            new Mechanic(39645, "Breakbar End", Mechanic.MechType.EnemyCastEnd, ParseEnum.BossIDS.Arkk, "symbol:'diamond-tall',color:'rgb(255,0,0)',", "CC.Fail","Breakbar (Failed CC)", "CC Fail",0,(condition => condition.getCombatItem().Value > 9668)),
+            new Mechanic(39645, "Breakbar End", Mechanic.MechType.EnemyCastEnd, ParseEnum.BossIDS.Arkk, "symbol:'diamond-tall',color:'rgb(0,160,0)',", "CCed","Breakbar broken", "CCed",0,(condition => condition.getCombatItem().Value < 9668)),
```
EliphasNUIT: shouldn't that be <= 9668 ? or does 9668 mean something else?

EliphasNUIT:
I was looking through several CC fail CC success mechanics. For a lot of them it's <= limit -> success, > limit -> fail. If it's the limit shouldn't we consider it as failure if that value is reached ? aka < limit -> success and >= limit -> fail ?

LunusLupus:
The issue is, that the game isn't precise with these and I am actually not sure if it fails when the value is reached or when the value is higher. Most buff removes are considerably later than the actual value so well never be the exact buff value (but at least 100 higher). And in general, successful events will result in lower values. 
I have seen Matthias end the sacrifice up to 25ms early, that's why I build in some buffer for that. In the end, we won't always be accurate but i guess people will understand that a break end with an immediate death (no downed state) was probably a false negative and a failed CC that ended without the person dying a false positive. We can ofc make it more strict and only produce one of the two errors at the cost of producing more errors overall.

EliphasNUIT:
I understand that for "variable" stuff like Matthias sacrifice and samarog brutalized. However for sloth, SH, demos stuff, aren't the values constant? For them I think we can accept reaching the value as failure (if I'm not wrong on their duration being constant ofc)

LunusLupus:
Their remove value is also always larger than the buff duration (because the removal event occurs later) when CC failed if I remember correctly, and lower when broken. As I said, I don't know if you can break them at the exact value or if it has to occur a ms before. I think we can decide for one scenario and make sure they are all consistent.

EliphasNUIT:
OK let's roll with that!

PR #154, opened 2018-09-12
Added log dir watcher and auto-parsing

Flomix:
I added 2 new settings, "Automatically parse added files" and "Automatically add new logs" (both default to false)
When the first is on parsing starts as soon as files are dropped in, when the 2nd is on every new log file (*.zip or *.evtc) in the gw2 log dir gets added to the list.
Files are added with 3 seconds delay, so the file is completely written and closed (or deleted) when handled.


EliphasNUIT:
Looks good! Thanks!

PR #158, opened 2018-09-13
Combat replay (Rectangular Actor)

LunusLupus:
- Added Rectangular Actor class
- Modified Surging Souls (Walls) at Soulless Horror
- Adjusted KC map

(Structure for custom icons for walls is done but commented out since it requires to have a look at the best implementation for custom sized replay icons. Only placeholder .pngs linked there as well.)

### DIFF
--- a/LuckParser/Models/ParseModels/CombatReplay/Actors/RectangleActor.cs
+++ b/LuckParser/Models/ParseModels/CombatReplay/Actors/RectangleActor.cs
@@ -7,19 +7,13 @@ namespace LuckParser.Models.ParseModels
         private readonly int _height;
         private readonly int _width;
 
-        public RectangleActor(bool fill, int growing, int height, int width, Tuple<int, int> lifespan, string color) : base(true, growing, lifespan, color, new MobileActor())
+        public RectangleActor(bool fill, int growing, int width, int height, Tuple<int, int> lifespan, string color) : base(true, growing, lifespan, color, new MobileActor())
         {
             _height = height;
             _width = width;
         }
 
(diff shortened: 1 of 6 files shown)


LuckParser/Models/ParseModels/CombatReplay/Actors/RectangleActor.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+
+namespace LuckParser.Models.ParseModels
+{
+    public class RectangleActor : Actor
+    {
+        private readonly int _height;
```
EliphasNUIT: I know I haven't gone through the fields for combat replay stuff yet, but this is better being a public readonly Height; or public Height {get;}, same for width

LuckParser/Models/ParseModels/CombatReplay/CombatReplay.cs
```diff
@@ -25,6 +25,7 @@ public class CombatReplay
         // actors
         private readonly List<CircleActor> _circleActors = new List<CircleActor>();
         private readonly List<DoughnutActor> _doughnutActors = new List<DoughnutActor>();
+        private readonly List<RectangleActor> _rectangleActors = new List<RectangleActor>();
```
EliphasNUIT: same remark as the fields for rectangle actor (no need to touch the rest, I'll do it once I get back to CR stuff)

LuckParser/Models/ParseModels/CombatReplay/Actors/RectangleActor.cs
LunusLupus: True, I did it this way for consistency with the other classes, but I'll change it if you plan it for the others as well anyways.

PR #164, opened 2018-09-14
Combat replay

LunusLupus:
- Included parsing of rotation values
- Added PieActor
- Added Soulless Horror's Slice Attacks, Dhuum's Cone pull and Deimos' Annihilate to the Combat Replay

### DIFF
--- a/LuckParser/Models/BossLogic/Deimos.cs
+++ b/LuckParser/Models/BossLogic/Deimos.cs
@@ -139,6 +139,25 @@ namespace LuckParser.Models
                     replay.AddCircleActor(new CircleActor(true, 0, 180, new Tuple<int, int>(start, end), "rgba(0, 0, 255, 0.3)", new Point3D(-8421.818f, 3091.72949f, -9.818082e8f, 216)));
                 }
             }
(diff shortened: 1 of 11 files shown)


LuckParser/Models/BossLogic/Deimos.cs
```diff
@@ -139,6 +139,25 @@ public override List<ParseEnum.TrashIDS> GetAdditionalData(CombatReplay replay,
...
                 }
             }
+            List<CastLog> annihilate = cls.Where(x => (x.SkillId == 38208) || (x.SkillId == 37929)).ToList();
+            foreach (CastLog c in annihilate)
+            {
+                int start = (int)c.Time;
+                int delay = 1000;
+                int end = start + 2400;
+                int duration = 120;
+                Point3D facing = replay.Rotations.FirstOrDefault(x => x.Time >= start);
```
EliphasNUIT: You should always check " != null" after a xxOrDefault. This will crash on logs done between position addition and facing addition to the evtc

LunusLupus: Ah, I assumed this was handled by the default part, my bad.

PR #176, opened 2018-09-20
Fix: Changed enemy mechanic table ordering to name column

Flomix:


EliphasNUIT:
The issue is fixed, but why are we showing boss mechanics table when there are no boss mechanics?
empty

Flomix:
Boss mechanic table is not created now when no mechanics are present.

PR #189, opened 2018-09-25
Maintenance

EliphasNUIT:
- Added condition cleansing to boon simulator, as a side effect conditions are also shown in boon plots
- Added automatic versioning:
    - Major and Minor are manual
    - Build -> number of days since last minor/major bump
    - Revision -> number of seconds since midnight divided by 2

__new html stuff__:
- Added condi plots to boon plots
- Added avg number of conditions, stacking condition uptimes to boss condition generation table 
- Added boss mechanics to dps plots


LuckParser/Resources/tmplBoonTable.html
```diff
@@ -14,7 +14,7 @@
 		<tr>
 			<td>{{:player.group}}</td>
 			<td title="{{:player.profession}}"><img src="{{:player.icon}}" alt="{{:player.profession}}" class="icon"><span style="display:none">{{:player.profession}}</span></td>
-			<td class="text-left"{{if !~generation}} title="Average number of boons: {{:data.avg}}"{{/if}}>{{:player.name}}</td>
+			<td class="text-center"{{if !~generation}} title="Average number of {{if ~condition}}conditions{{else}}boons{{/if}}: {{:data.avg}}"{{/if}}>{{:player.name}}</td>
```
Flomix: Hmmm I would stick with a left-aligned name column. Other tables have that too, and especially with the icon column in the left it looks cleaner when the icon is always directly next to the player name.

EliphasNUIT: The reason I ended up putting it was because the tool tips were not aligned with the text, they appear from the center while the names are on the left and it was kinda bothering me.

Flomix: Still, the layout of the table is more important than the layout of the tooltips?

PR #201, opened 2018-10-02
Added new phase-mode chart

Flomix:
Add new chart mode "Phase".
In the full fight chart the dps counter is reset at every phase transition; the result is basically the same as if the single phase graphs are combined to a single graph; that makes phases comparable.

The phase graph is available in phase tabs too (for technical reasons), there they behave exactly as the "Full" graph.

### DIFF
--- a/LuckParser/Controllers/HTMLBuilderNew.cs
+++ b/LuckParser/Controllers/HTMLBuilderNew.cs
@@ -1523,11 +1523,14 @@ namespace LuckParser.Controllers
             data.graphs.Add(new GraphDto("full", "Full"));
             data.graphs.Add(new GraphDto("s10", "10s"));
             data.graphs.Add(new GraphDto("s30", "30s"));
+            data.graphs.Add(new GraphDto("phase", "Phase"));
 
             for (int i = 0; i < _statistics.Phases.Count; i++)
             {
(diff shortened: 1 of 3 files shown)


EliphasNUIT:
Regarding that technical reasons, could you just not add "phase" attributes in extractDpsData and to totDps when i > 0 (aka we are not in FullFight phase)? That would also require giving phaseIndex to extractDpsData and checking it. I'm kinda bothered that we are showing stuff in phases that does not mean anything/shouldn't even be there to begin with.
Unless I'm missing something obvious?

Flomix:
I can try to improve it at that point. Right now there is no logic, I have a list of graph styles (like "Full", "s30" etc) and I have a list of phases, and every graph style is offered for every phase.
The same is true for the s30 graph for a phase that is only 20s short; that too doesn't make sense because it is identical to "Full"

EliphasNUIT:
True, but the difference is your dev has been specifically done for the "Full Fight" phase as its aim is to merge the graphs of the different phases. It's not on a "is this usefull" level I'm speaking (like your 30s graph example in a 20s phase) but on a conceptual level, the "phase" graph should not appear in phases, useful or not.

Flomix:
PR modified

PR #212, opened 2018-10-11
Settings rework

Flomix:
- Settings are not lost anymore on a version change
- the `-c` parameter for custom settings can now be used for UI mode too
- Rewrote how custom settings are handled
    - "old" `*.config` files in xml format still work for backwards compatibility (a deprecated-message is printed to the console). The known errors are still present there.
    - new config file extension `*.conf`
    - file uses a simple `key=value` format similar to many linux programs
    - comment lines start with `#` and are ignored
    - a `sample.conf` with a sample config is shipped along with the .exe
    - invalid lines, or unknown keys are ignored with a message to the console, it is not possible anymore that the program crashes
    - The priority of the settings was changed to custom file > user settings > shipped defaults:
        - Settings in the custom file have the highest priority
        - If a key is not present in the custom file or no custom file is used, user settings created by the settings dialog kick in
        - If no user are present or a key is not present there the app defaults kick in

### DIFF
--- a/LuckParser/App.config
+++ b/LuckParser/App.config
@@ -112,6 +112,9 @@
             <setting name="AutoAddPath" serializeAs="String">
                 <value />
             </setting>
+            <setting name="Outdated" serializeAs="String">
+                <value>True</value>
+            </setting>
         </LuckParser.Properties.Settings>
     </userSettings>
 </configuration>
--- a/LuckParser/ConsoleProgram.cs
+++ b/LuckParser/ConsoleProgram.cs
@@ -11,11 +11,11 @@ namespace LuckParser
 {
     public class ConsoleProgram
     {
-        public ConsoleProgram(string[] args)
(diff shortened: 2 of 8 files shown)


Flomix:
A settings file can be as simple as
```
#
# My wonderful Configuration for HTML files
#
PlayerRot=true
SimpleRotation=true
SaveOutHTML=true
NewHtmlMode=true
NewHtmlExternalScripts=false
LightTheme=true
ParseCombatReplay=true

# Add files as they are created, and parse them automatically:
AutoAdd=true
AutoParse=true
AutoAddPath=C:\Users\me\Documents\Guild Wars 2\addons\arcdps\arcdps.cbtlogs
```

Flomix:
the Outdated is used to migrate the previous settings on version change, see Program.js:51. That way I can see if the settings are defaults or not; "Outdated" defaults to true and I set it to false in code.

PR #220, opened 2018-10-18
Separate static and dynamic combat replay script

Flomix:
- Moved all code in the combatreplay.js into functions
- changed all dynamic placeholders in combatreplay.js to function arguments
- Split the generated script to 2 script blocks, one static part and a generated part, that simply calls a static init function with replay data as arguments

With these changes it is a trivial step to make the combatreplay.js a cacheable external resource, like the other javascripts too

### DIFF
--- a/LuckParser/Controllers/HTMLHelper.cs
+++ b/LuckParser/Controllers/HTMLHelper.cs
@@ -5,6 +5,7 @@ using System.Drawing;
 using System.IO;
 using System.Linq;
 using LuckParser.Models.DataModels;
+using Newtonsoft.Json;
 
 namespace LuckParser.Controllers
 {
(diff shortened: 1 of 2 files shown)


LuckParser/Controllers/HTMLHelper.cs
```diff
@@ -749,69 +750,74 @@ public static void WriteCombatReplayInterface(StreamWriter sw, Tuple<int,int> ca
...
+            string actors = "";
+            int count = 0;
+            foreach (Player p in log.PlayerList)
             {
-                string replayScript = Properties.Resources.combatreplay_js;
-                replayScript = replayScript.Replace("'${inch}'", map.GetInch().ToString());
-                replayScript = replayScript.Replace("'${pollingRate}'", pollingRate.ToString());
-                replayScript = replayScript.Replace("'${mapLink}'", map.Link);
-                string actors = "";
-                CombatReplay replay;
```
Flomix: most of the following lines is just whitespace changes, you should click "**Diff settings**" and enable "**Hide whitespace changes**" here.

PR #233, opened 2018-10-31
Boons: fractal potions and mistlock singularity

cordbleibaum:


EliphasNUIT:
The potions should be consumables, not standard buffs.

cordbleibaum:
ok, switched pots consumable, there is still the issue tho that there are apparently no stackable consumables yet

EliphasNUIT:
It'll be properly displayed on the graph. Shouldn't be difficult to adapt GetConsumables() on Player.cs to take that into account

cordbleibaum:
the problem is in the player tab every consumable stack gets displayed as a seperate consumable making it hard to read

LunusLupus:
image

I don't see any issue here.

EliphasNUIT:
if everything is applied at the same time, we can easily do consumable (stacks) when applicable, if someone takes during a fight potions 1 by 1, it'll appear that number of time in refreshed

EliphasNUIT:
I'll merge this this evening, and add the stacking support to that part

PR #320, opened 2019-01-20
Phases

EliphasNUIT:
- Added subphases to Dhuum's main phase separated by greater deathmarks + cata cycles
- Added phases to mursaat base on hp
- Added phases to slothasor based on breakbar
- Added phases to soulless horror based on breakbar

### DIFF
--- a/LuckParser/Models/FightLogic/Dhuum.cs
+++ b/LuckParser/Models/FightLogic/Dhuum.cs
@@ -61,6 +61,41 @@ namespace LuckParser.Models.Logic
             }
         }
 
+        private List<PhaseData> GetInBetweenSoulSplits(ParsedLog log, Target dhuum, long mainStart, long mainEnd)
+        {
+            List<CastLog> cls = dhuum.GetCastLogs(log, 0, log.FightData.FightDuration);
+            List<CastLog> cataCycle = cls.Where(x => x.SkillId == 48398).ToList();
+            List<CastLog> gDeathmark = cls.Where(x => x.SkillId == 48210).ToList();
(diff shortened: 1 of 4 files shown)


Sejsel:
I am not sure the MO phases provide any useful data, I don't think it's necessary to have some kind of phases on each encounter.

Slothasor phases seem very useful, SH will have a variable amount of phases, but that shouldn't be an issue or confusing in any way.

The Dhuum phases are overlapping, right? At least that is what I am guessing by names. That might be a bit misleading if somebody thinks of phases as non-overlapping which was the case in EI for a long time.

EliphasNUIT:
EI had overlapping phases since the very first day phases were implemented, with KC (burst phases included inside their respective main phases) and Deimos add phases have existed for at least half a year now which are basically sub phases of the 100% - 10% phase focused on Sauls. Largos also has overlapping phases by nature.  You could also argue that all the phases overlap with the full fight phase :P
For MO I've just put it because it was possible, some may found some use to it or not, does not cost to have it. That'll also be helpful if some day we animate the arena with the spear pattern.

Sejsel:
Oh, yeah, I remember somebody being confused about KC phases, in that case it's not very well communicated. You are right about Deimos, forgot about that one. I guess Sabetha works similarly. If it's communicated well, like naming the phase pre-something as in this case, it's fine.

And it's true that the phases would be useful for the spear pattern, so I guess it's fine.

EliphasNUIT:
Yeah I had changed some naming after the KC confusion and added pre-burn phases and leftover phases to it

PR #323, opened 2019-01-26
Add moving platforms to Qadim combat replay

Sejsel:
Adds support for background actors for combat replay and implements moving platforms that interpolate between positions in time with support for rotations and opacity changes.
These might be useful in the future, but could, for example, also be used in the Twin Largos encounter to show height of platforms through opacity changes.

**If this gets merged in the current state, the remaining issues have to be added to issues.**

### Qadim

Adds moving platforms for the whole fights with timings determined from phase data. If it is missing (perhaps because it's not calculated), only the first position will be used. For the very last move, Qadim's position is checked instead because there is no phase for that.

Due to limitations of the combat replay as it's written now where each entity has its own combat replay, moving platforms are added to Qadim's combat replay.

You can see how it looks [here](https://sejsel.cz/tmp/ei/qadim-2019-01-26_18.26.00.webm) (6.5 MB video file).

#### Caveats
- The method that adds platforms is rather... long. This is because I opted for defining reused coordinates as constants to avoid mistakes from only editing a coordinate in one place when it's actually used in multiple.
- This currently depends on phase definitions not changing (including the full fight phase) as that would break timings. A warning was added to the `GetPhases()` method.

#### Remaining Issues
- Sometimes some actors do not render you open the combat replay and did not press play yet. It is not a lifetime issue. It can happen to platforms, it can happen to player icons, but at the same time it might not happen to their facings. It's weird, I have no clue what it's caused by.

- Either I am doing some math wrong or there is something very weird about drawing images, almost as if there was a transform that'd make coordinates of everything further apart in the x axis, but not y.

After spending quite a lot of time on this, there are two possible outputs. Prior to the change in #c871f38, there were vertical gaps between platforms

Before

In #c871f38, the platform image was changed to properly fill between the corners, but the image is no longer an equilateral triangle (although the vertices do form one in the game!). This looks somewhat better until the jumping puzzle phase where platforms rotate and it becomes obvious.

After the change

In the jumping puzzle

It might be fixable by modifying the texture somehow, but I do not know how, feel free to fix this (please do!)

- Qadim combat replay performance is a bit worse than combat replay of other encounters, presumably due to rendering 12 more images each frame. Mostly noticeable on Firefox. Might look into that in the future. When platforms are not moving they wouldn't need to be rendered separately, but there may be better ways to improve performance.

#### To add in the (soon) future
- The blue spheres for jumping between platforms at Qadim. A simple addition, just want to get this done first.
- Some coordinates of hidden platforms are not correct. Putting a marker on them and recording a run should be enough to find where all of them are. All these positions are marked with a `// TODO` comment.
- Heights for platforms in the wyvern phase are not currently correct, but it's irrelevant for combat replay as it's now as no platforms overlap there.

### DIFF
--- a/LuckParser/LuckParser.csproj
+++ b/LuckParser/LuckParser.csproj
@@ -113,6 +113,7 @@
     <Compile Include="Models\ParseModels\CombatReplay\Actors\BackgroundActor.cs" />
     <Compile Include="Models\ParseModels\CombatReplay\Actors\FacingActor.cs" />
     <Compile Include="Models\ParseModels\CombatReplay\Actors\GenericActor.cs" />
+    <Compile Include="Models\ParseModels\CombatReplay\Actors\MovingPlatformActor.cs" />
     <Compile Include="Models\ParseModels\Logs\BoonExtensionLog.cs" />
     <Compile Include="Models\ParseModels\Simulator\BoonSimulationItems\AbstractBoonSimulationItem.cs" />
     <Compile Include="Models\ParseModels\Simulator\BoonSimulationItems\BoonSimulationItemOverstack.cs" />
--- a/LuckParser/LuckParser.csproj
+++ b/LuckParser/LuckParser.csproj
@@ -110,6 +110,7 @@
     <Compile Include="Models\FightLogic\BanditTrio.cs" />
     <Compile Include="Models\FightLogic\Freezie.cs" />
     <Compile Include="Models\JsonModels\JsonDeathRecap.cs" />
+    <Compile Include="Models\ParseModels\CombatReplay\Actors\BackgroundActor.cs" />
     <Compile Include="Models\ParseModels\CombatReplay\Actors\FacingActor.cs" />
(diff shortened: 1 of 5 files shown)


EliphasNUIT:
That's some good stuff here!

Regarding your issues:
- For the missing images without "play", I believe this comes from the fact that at that given time the images were not loaded yet. As of today I only do an image.onLoad for the background image because it was always the biggest image, with your dev, the background image should be much smaller. A simple fix would be adding onLoad to each image
- I don't have anything to say for your "shifted" issue for now, first time I'm hearing something like that. However my first guess would be the "rect" attribute of the map is not perfect. That part, since W5, is done by putting people/markers on key places + manual corrections, as those maps are not in the official tile API. We just come to accept small innacuracies.
- I'll check performances sooner or later, once I'm done with my other devs and get back to CR. I was thinking of a two layer canvas, on one we only draw static actors (once, or when the viewpoint changes) and the other for dynamic stuff. I'll still check the javascript code in its totality, maybe we have deopts here and there impacting performances. Just to be sure, hardware acceleration was enabled on your browsers right?

Regardless, thank you.

PR #335, opened 2019-02-09
added skills from downed and transformed states to AA flag

greaka:
The downed state and also some transformations, such as holoforge, pets or shroud are marked as DOWNED_1 instead of WEAPON_1.

This pull requests aims to add them to the AA flag to have them hidden when the simple rotation flag "show auto attacks" is deactivated.


EliphasNUIT:
Thanks

PR #419, opened 2019-07-21
Maintenance

EliphasNUIT:
- Removed some more erroneous success/failure detections
- Fixed group/squad uptimes data on phase active time
- Added phase active time attribute to json output

### DIFF
--- a/LuckParser/FightLogic/BanditTrio.cs
+++ b/LuckParser/FightLogic/BanditTrio.cs
@@ -70,20 +70,6 @@ namespace LuckParser.Logic
             }
         }
 
-        public override void SpecialParse(FightData fightData, AgentData agentData, List<CombatItem> combatData)
-        {
-            ComputeFightTargets(agentData, fightData, combatData);
-            Target narella = Targets.Find(x => x.ID == (ushort)ParseEnum.TargetIDS.Narella);
-            if (narella != null)
-            {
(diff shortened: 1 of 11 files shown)


LuckParser/FightLogic/BanditTrio.cs
```diff
@@ -70,6 +70,20 @@ public override void CheckSuccess(CombatData combatData, AgentData agentData, Fi
...
+        public override void SpecialParse(FightData fightData, AgentData agentData, List<CombatItem> combatData)
+        {
+            ComputeFightTargets(agentData, fightData, combatData);
+            Target narella = Targets.Find(x => x.ID == (ushort)ParseEnum.TargetIDS.Narella);
+            if (narella != null)
+            {
+                CombatItem c = combatData.FirstOrDefault(x => x.IsStateChange == ParseEnum.StateChange.ExitCombat && x.SrcInstid == narella.InstID && x.LogTime <= narella.LastAwareLogTime && x.LogTime >= narella.FirstAwareLogTime);
+                if (c != null)
+                {
+                    combatData.Remove(c);
```
Sejsel: I don't believe this will always work. I think Narella can enter combat multiple times if players hit her with something while she's on the balcony. Also, this will could cause some weird behavior if you expect that you can never enter combat without leaving first.

Because this is used for the success checking, I believe all that needs to be done is checking if the last exit is not followed with another combat enter.

EliphasNUIT: Fair point, I have changed the algorithm to check for a combat exit after the last combat enter.

PR #462, opened 2020-04-28
Consumables

Flomix:
Added (all?) missing utility and ascended food consumables.

Many ascended foods don't have an ID in the wiki yet, they were added nonetheless without an ID to make later completion easier and to track, what is missing.


EliphasNUIT:
Looks great, thanks!

PR #486, opened 2020-08-21
Instant cast

EliphasNUIT:
Initial instant cast detection support


hristov296:
Is it possible to also have the mesmer instant casts?

EliphasNUIT:
I'll be gradually adding everything that can be reasonably added. For now I have only added simple stuff and the framework for instant cast detection

PR #492, opened 2020-09-14
Add stopPropagation to CR Mech list filter clicks

hobinjk:
Prevents the dropdown from closing every time a filter checkbox label was clicked.


EliphasNUIT:
Thanks!

PR #495, opened 2020-09-16
adjusting Skorvald CM detection

networkjanitor:
Since Shattered Observatory moved to 99 (from 100), all the HP values for the bosses changed.


EliphasNUIT:
This breaks older log parses though, I'm changing == with >=.
Thanks!

PR #513, opened 2020-12-12
Cairn stuff

EliphasNUIT:
- Added enrage phases for Cairn
- Noticed the "spawn protection check" was not always working when logs start sooner than expected as the buff remove event is missing, as such, implemented two fallbacks to this method, both using first Impact(38102) cast end event. From the skill dump, the effect of the cast happens at 1025 ms:
   - If scaled duration is present, use the scaled/nonScaled ratio to transform 1025. Get the fight offset by doing logStart - actualDuration + effectHappened
   - If scaled duration is not present (old logs), simply do logStart - actualDuration + 1025

From my measurements, the first fallback adds between 10ms  and 15 ms to the fight while the 2nd adds 15ms to 20ms. The difference shouldn't be big as Cairn never starts with quickness.
- Also added AnimatedCastEvents for when the first seen activation is an EndCasting(). As such skills which started before log start but ended after it will appear on your rotation. It'll have proper acceleration and saved time values (in CANCEL_FIRE situations) with logs that have scaled durations.


Sejsel:
Thank you for the enrage phases, I really appreciate it!

PR #538, opened 2021-01-24
Corrected target health for Deimos NM 10% and refactored GetPhases by…

### LINKED ISSUE #537: Deimos displaying CM health on normal mode <10%
**Describe the bug**
In logs parsed with the current version Deimos NM health always shows as 42804900 if the 10% phase was reached. 
In logs using version 2.23 normal mode deimos shows as 37388210 instead if the 10% phase was reached which should be the correct value.

**To Reproduce**
Should be reproducable with any Deimos NM log that reaches 10%. Parse attached log as a sample.
[20210122-221015.zip](https://github.com/baaron4/GW2-Elite-Insights-Parser/files/5862948/20210122-221015.zip)
Output on dps.report: https://dps.report/nd0l-20210122-221012_dei

**Cause**
This was likely caused by [this commit](https://github.com/baaron4/GW2-Elite-Insights-Parser/commit/d8c173ff7cf31a5ef7d258ab3c76dee7d458b735) which changed CM evaluation from a bool-like int to an enum but didn't change the evaluation of `res > 0` along with the introduction of the enum, causing any Deimos to reach 10% to always apply the CM 10% health value.

Endaris:
Fixes #537 

I found that changing the health inside a method that only checks a state (IsCM) was quite counterintuitive so I was looking to move it into GetPhases instead. The code for that was a bit difficult to understand with the changes to the local start+end variables so I refactored the method to get rid of these with the goal of making it easier to read and ended up creating some private methods for easier readability.

### DIFF
--- a/GW2EIEvtcParser/EncounterLogic/Raids/W4/Deimos.cs
+++ b/GW2EIEvtcParser/EncounterLogic/Raids/W4/Deimos.cs
@@ -307,24 +307,11 @@ namespace GW2EIEvtcParser.EncounterLogic
                 phase100to10.AddTarget(mainTarget);
                 phases.Add(phase100to10);
 
-                if (_deimos10PercentTime > 0)
+                if (_deimos10PercentTime > 0 && log.FightData.FightEnd - _deimos10PercentTime > ParserHelper.PhaseTimeLimit)
                 {
-                    // Deimos gains additional health during the last 10% so the max-health needs to be corrected
-                    if (IsCM(log.CombatData, log.AgentData, log.FightData) == FightData.CMStatus.CM)
-                    {
-                        mainTarget.SetManualHealth(42804900);
-                    }
-                    else
-                    {
-                        mainTarget.SetManualHealth(37388210);
-                    }
-
-                    if (log.FightData.FightEnd - _deimos10PercentTime > ParserHelper.PhaseTimeLimit)
-                    {
-                        var phase10to0 = new PhaseData(_deimos10PercentTime, log.FightData.FightEnd, "10% - 0%");
-                        phase10to0.AddTarget(mainTarget);
-                        phases.Add(phase10to0);
-                    }
+                    var phase10to0 = new PhaseData(_deimos10PercentTime, log.FightData.FightEnd, "10% - 0%");
+                    phase10to0.AddTarget(mainTarget);
+                    phases.Add(phase10to0);
                 }
                 //mainTarget.AddCustomCastLog(end, -6, (int)(start - end), ParseEnum.Activation.None, (int)(start - end), ParseEnum.Activation.None, log);
             }
@@ -514,7 +501,23 @@ namespace GW2EIEvtcParser.EncounterLogic
             {
                 throw new MissingKeyActorsException("Deimos not found");
             }
-            return (target.GetHealth(combatData) > 40e6) ? FightData.CMStatus.CM : FightData.CMStatus.NoCM;
+            FightData.CMStatus cmStatus = (target.GetHealth(combatData) > 40e6) ? FightData.CMStatus.CM : FightData.CMStatus.NoCM;
+
+            if (_deimos10PercentTime > 0)
+            {
+                // Deimos gains additional health during the last 10% so the max-health needs to be corrected
+                // done here because this method will get called during the creation of the ParsedEvtcLog and the ParsedEvtcLog should contain complete and correct values after creation
+                if (cmStatus == FightData.CMStatus.CM)
+                {
+                    target.SetManualHealth(42804900);
+                }
+                else
+                {
+                    target.SetManualHealth(37388210);
+                }
+            }
+
+            return cmStatus;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


GW2EIEvtcParser/EncounterLogic/Raids/W4/Deimos.cs
```diff
@@ -278,59 +278,83 @@ internal override void EIEvtcParse(FightData fightData, AgentData agentData, Lis
...
             if (invulDei != null)
             {
-                end = invulDei.Time;
-                phases.Add(new PhaseData(start, end));
-                start = _deimos10PercentTime > 0 ? _deimos10PercentTime : fightDuration;
+                var phase100to10 = new PhaseData(0, invulDei.Time, "100% - 10%");
+                phase100to10.AddTarget(mainTarget);
+                phases.Add(phase100to10);
+
+                if (_deimos10PercentTime > 0)
```
EliphasNUIT, line 310: If phases are not parsed, the health is never corrected.

Also the reasons why the hp modification was done in check CM were:
- Modifying the content of the agent/combat event outside of the EvtcParsedLog contructor is very bad practice, it is not necessary to call getPhases(...) so that would mean the log is in a bad state (event without counting the requirePhases conditioning)
- The modification only depends on if deimos10% was reached and isCM, both known after EIEvtcParse and IsCM respectively and both are called before EvtcParsedLog is created

I like the overall refactoring, that is always appreciated, but I very much prefer keeping "no agent or combat event modification done" once ParsedEvtcLog has been created as EIParser can be plugged into any other application for parsing purposes and I'm not in favor of possibly having non complete logs unless some public APIs have been called post constructor.

Endaris: You're right about health not being corrected if phases are not being parsed, I overlooked that use-case because I never turn phases off.

Regarding the comments about bad state, I can't fully comprehend the architectural implications behind it yet but I'll just believe you that it's bad.
I guess I was going too much along what I had seen for the GetPhases on Xera for correcting the health.
Wouldn't setting the health for Xera's 2nd phase cause a similar problem?

I looked up where the `IsCM` method gets called and found that it's called in the `SetCM` method of FightData.
I think that splitting the content of that method into 2 methods `IsCm` (only fetching the state) and a new interface method `ApplyCmChanges` (actually changing the health value) would make it more intuitive but given that Deimos seems to be the only use case for this, it's probably not a significant addition. 
For now I'll move it back into IsCM and add a comment with the reasoning why it's there.

Thanks for the feedback, I'll look to get it done in the next half hour.

EliphasNUIT: To give you an example, if we do it in GetPhases(). Your user will parse their log, then, for example, look at the targets list in order to find Deimos. They will then call GetHealth() on it in order to check its health for one reason or another. As that user has not called GetPhases() before hand, they'll get the non corrected health value and unless they call GetHealth() a second time after calling GetPhases() (not sure they'll even call it), they'll be stuck with a bad value. 
If we do it inside the ParsedEvtcLog constructor, where all that CM and specific EI processing/redirection stuff happens, then your user will never have to worry about the order in which they need to call public APIs to make sure the values they get are correct.

You are absolutely correct with your 2nd point. The code is pretty big, I do a lot of cleanup and refactoring but sometimes stuff gets past the net. I'll do a pass on name/hp override related stuff to make sure they are called in correct spots.

Regarding the 3rd, yes, that could be done. I did not particularly bother with it because it is very specific to Deimos, for the others it can always be done in EIEvtcParse where EncounterLogic specific operations are done. That would definitively be the path to take should additional encounters with such situations arise or if there is a need to do much more complex operations around CM conditions.

GW2EIEvtcParser/EncounterLogic/Raids/W4/Deimos.cs
```diff
@@ -490,12 +501,23 @@ internal override FightData.CMStatus IsCM(CombatData combatData, AgentData agent
...
                 throw new MissingKeyActorsException("Deimos not found");
             }
-            FightData.CMStatus res = (target.GetHealth(combatData) > 40e6) ? FightData.CMStatus.CM : FightData.CMStatus.NoCM;
+            FightData.CMStatus cmStatus = (target.GetHealth(combatData) > 40e6) ? FightData.CMStatus.CM : FightData.CMStatus.NoCM;
+
             if (_deimos10PercentTime > 0)
             {
-                target.SetManualHealth(res > 0 ? 42804900 : 37388210);
+                // Deimos gains additional health during the last 10% so the max-health needs to be corrected
+                // done here because this method will get called during the creation of the ParsedEvtcLog and the ParsedEvtcLog should contain complete and correct values after creation
```
EliphasNUIT, line 509: Thanks!
I'll take a more detailled look at the full PR by tomorrow.

GW2EIEvtcParser/EncounterLogic/Raids/W4/Deimos.cs
Endaris: Thank you for the explanation :-)

EliphasNUIT:
Looks good! Thanks!

Issue #520, opened 2020-12-20
Allow external/CDN usage for js & css


aeris:
Currently GW2EIP use hardcoded `./EliteInsights-2.28.0.js` & `./EliteInsights-CR-2.28.0.js` path for resources.
It could be cool to have an option to specify the expected path for those resources and so allowing shared path like with a CDN (`https://cdn.example.org/EliteInsights-CR-2.28.0.js`) or logs in subfolders using a common absolute path (`/assets/EliteInsights-2.28.0.js`).

EliphasNUIT:
I'll add this to my backlog but this is going to be very low priority compared to other tasks I've planified.

EliphasNUIT:
Added in #540 thanks to @tschoeck

aeris:
ohhhhhhhhhhhh :heart_eyes: 
Thanks a lot @tschoeck

PR #540, opened 2021-01-27
Custom source directories

tschoeck:
I added a feature based on #520 

You have now two more options on the settings tab "HTML". Both are only available if you activated "External scripts"
image

1) Fill in the name of a subdirectory here to place external script & css files in a subdirectory based on the root dir of the report. Remeber to use valid chars here otherwise this option will be ignored. The html report file will use this directory to include those files.. except if you fill an url in option 2:

2) Fill in an Url. The report html file will use this url to include css and script files. They have to be saved there once.
- Remember to think about CORS if you use a seperate server for holding the external files
- You have to save the external files on the url location once. The reports will not work if you forget this. And if you patch the version of the tool you have to upload those static files once again.

### DIFF
--- a/GW2EIParser/Settings/SettingsForm.Designer.cs
+++ b/GW2EIParser/Settings/SettingsForm.Designer.cs
@@ -61,6 +61,8 @@
             this.ChkAnonymous = new System.Windows.Forms.CheckBox();
             this.ChkHtmlExternalScripts = new System.Windows.Forms.CheckBox();
             this.ChkDetailledWvW = new System.Windows.Forms.CheckBox();
+            this.LblHtmlExternalScriptsCdn = new System.Windows.Forms.Label();
+            this.LblHtmlExternalScriptsPath = new System.Windows.Forms.Label();
             this.TxtHtmlExternalScriptsCdn = new System.Windows.Forms.TextBox();
             this.TxtHtmlExternalScriptsPath = new System.Windows.Forms.TextBox();
             this.ChkSaveOutTrace = new System.Windows.Forms.CheckBox();
@@ -79,8 +81,6 @@
             this.TabHTML = new System.Windows.Forms.TabPage();
             this.PictureTheme = new System.Windows.Forms.PictureBox();
             this.PanelHtml = new System.Windows.Forms.Panel();
-            this.LblHtmlExternalScriptsCdn = new System.Windows.Forms.Label();
-            this.LblHtmlExternalScriptsPath = new System.Windows.Forms.Label();
             this.PanelTheme = new System.Windows.Forms.Panel();
             this.RadioThemeLight = new System.Windows.Forms.RadioButton();
             this.RadioThemeDark = new System.Windows.Forms.RadioButton();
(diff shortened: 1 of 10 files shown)


EliphasNUIT:
Thanks! I'll take a look.

GW2EIParser/Settings/SettingsForm.cs
```diff
@@ -38,6 +38,7 @@ public void ConditionalSettingDisable(bool busy)
             BtnResetTraitList.Enabled = !busy;
             BtnLoadSettings.Enabled = !busy;
             GroupWebhookSettings.Enabled = !busy;
+            TxtHtmlExternalScriptsPath.Enabled = !busy;
```
EliphasNUIT, line 41: TxtHtmlExtralScriptsCdn should also be enabled on !busy

GW2EIParser/Settings/SettingsForm.Designer.cs
```diff
@@ -449,10 +453,39 @@ private void InitializeComponent()
...
+        "r and the generated files bigger");
+            this.ChkDetailledWvW.UseVisualStyleBackColor = true;
+            this.ChkDetailledWvW.CheckedChanged += new System.EventHandler(this.ChkDetailledWvWCheckedChange);
+            // 
+            // TxtHtmlExternalScriptsCdn
+            // 
+            this.TxtHtmlExternalScriptsCdn.Location = new System.Drawing.Point(44, 54);
+            this.TxtHtmlExternalScriptsCdn.Name = "TxtHtmlExternalScriptsCdn";
+            this.TxtHtmlExternalScriptsCdn.Size = new System.Drawing.Size(157, 20);
+            this.TxtHtmlExternalScriptsCdn.TabIndex = 57;
```
EliphasNUIT, line 478: A tooltip here to indicate this has priority over ScriptsPath and also warn about CORS related errors would be nice

README.md
```diff
@@ -103,6 +103,10 @@ Note it may take some time for each file to parse and they will not be ready to
 
 -__HtmlExternalScripts__: if true, css and js files will be separated from the html.
 
+-__HtmlExternalScriptsPath__: Available if HtmlExternalScripts is enabled. Will place the external script files in a subdirectory.
```
EliphasNUIT, line 106: Readme should also indicate that this path is relative.

EliphasNUIT:
Perhaps it is better for the path to be absolute?

tschoeck:
Thank you for your fast review. I added your recommendations and set the path to work with absolute paths only. The settings ui was updated with a better name for the path option and tooltips.

I think the number of users, using this tool locally are quite higher than the users who use it on a server to populate a service (like dps.report). For server usage: if you set an absolute path and dont set a cdn path the chances are high, that this absolute path might not be accessable for your visitors (I have mentioned this in the readme). So if you use the option "Absolute Path" on a server, you have to set the cdn option too. On the other hand, you will not have to copy the external sources on the server yourself, you could just populate the absolute path via webserver and point to it via cdn url.

Should there be a button to select the path easily? Currently you have to copy or type it into the textbox yourself.

EliphasNUIT:
Everything looks good to me. I think it makes sense for server hosting to use cdn path option and for local users to use absolute path option.

A button to select the the path would be nice yes. If you have the time, feel free to add it but I don't see it as a mandatory addition.

Regardless, I'm ok with merging this.

tschoeck:
Thank you for your feedback. I added a select button behind absolute path. Feel free to merge if you like.

EliphasNUIT:
Thanks a lot!

PR #572, opened 2021-06-27
Qadim the Peerless CR

LunusLupus:
- Added the Force of Havoc mechanic to CR
- Added the small green circle around uncaught Energy Orbs in the CR
- Fixed Unbridled Chaos image file


EliphasNUIT:
Looks great, thanks!

PR #575, opened 2021-07-11
QtP CR+Mechanics; Color additions; Decoration adjustments

LunusLupus:
General:
- Added named Colors to streamline coloring of the Mechanics
- Adjusted FacingRectangleDecoration to allow Translation (into radial direction, same as RotatedRectangleDecoration)

QtP:
- Added Critical Mass (Orb caught) to Mechanics/Buffs
- Added Critical Mass (Orb caught) red circles to players in the CR
- Added Exponential Repercussion (Pylon/Qadim Dome Knockback) to Mechanics + CR
- Added Force of Retaliation (Large cascading shockwave) to CR
- Added Ether Strikes to CR
- Added Caustic Corrosion to CR
- Added Caustic Chaos to CR


GW2EIEvtcParser/EncounterLogic/Raids/W7/PeerlessQadim.cs
```diff
@@ -191,6 +197,72 @@ internal override void ComputeNPCCombatReplayActors(NPC target, ParsedEvtcLog lo
...
+                    foreach (AbstractCastEvent c in expoReperc)
+                    {
+                        int radius = 650;
+                        start = (int)c.Time;
+                        end = (int)c.EndTime;
+                        Point3D position = replay.Positions.LastOrDefault(x => x.Time <= start + 1000);
+                        replay.Decorations.Add(new CircleDecoration(true, 0, radius, (start, end), "rgba(255, 220, 0, 0.15)", new PositionConnector(position)));
+                        replay.Decorations.Add(new CircleDecoration(true, end, radius, (start, end), "rgba(255, 220, 50, 0.25)", new PositionConnector(position)));
+
+                        foreach (NPC pylon in TrashMobs.Where(x => x.ID == 21962))
```
EliphasNUIT, line 260: Wouldn't it make more sense for that to happen while processing pylons in ComputeNPCCombatReplayActors instead of QtP and put the circle decorations inside their combat replay instead of Qadim's?

LunusLupus: That was my initial idea as well.
The reason why I put it there is that the skill detection has to come through qadim's skill casts anyway since the pylons' skill casts aren't visible. Sadly, this means that the knockback when a pylon is unguarded will also not show up, but I didn't find a way to detect that.

EliphasNUIT: Ok, I see. I'll see if there could be a cleaner way to do it in the future, it feels a little bit hacky but I don't see any functional problems to it.

EliphasNUIT:
Looks good to me, thanks!

GW2EIEvtcParser/EncounterLogic/Raids/W7/PeerlessQadim.cs
LunusLupus: Indeed. I didn't like it too much considering its hacky nature but was the best I could do for the moment.

PR #634, opened 2022-03-24
Skill IDs

EliphasNUIT:
Created a static class that contains a big pool of skill ids.


EliphasNUIT:
First drop, migrated buff, damage modifiers and prof helpers

PR #635, opened 2022-03-25
Add basic support for the open world Soo-Won encounter

Stonos:
This PR adds basic support (targets & phases) for the open world Soo-Won encounter.

image

Soo-Won appears to have 3 different invulnerability buffs:
* **Invulnerability (buff ID `757`):** Applied between different phases.
* **Damage Immunity (buff ID `66242`):** Applied during the spear phases at 60% and 40%. The `757` buff gets removed when this is applied.
* **Damage Immunity (buff ID `65716`):** Not sure about this one, it seems to randomly get applied (maybe related to the tail?). This one is ignored.

Due to this, I have updated `GetPhasesByInvul` to make it accept a list of skill IDs instead of a single skill ID.


EliphasNUIT:
Looks great! I'll take a look with more details to what you did for the phases and the offsets just to understand it.
Regardless, I don't see any issue with the code.

EliphasNUIT:
Everything looks great. I'll keep the PR open for a few days though, I wish to quickly make a release with a few bug fixes + polish on 30th after the mini balance patch, I'll merge this after that.

EliphasNUIT:
Thanks!

PR #658, opened 2022-08-24
Hook up unlinked ascended food IDs

marcustyphoon:
I thought I could use a script to automatically add all of the ascended food ids before realizing that item IDs and buff IDs aren't the same. Oh, well, no matter.

This adds blank entries to SkillIDs.cs for the missing ascended foods so they can be entered with less effort, and fixes five foods whose IDs were already entered but were not referenced in Buff.cs: Spherified Cilantro Oyster Soup, Strawberry Cilantro Cheesecake, Coq Au Vin with Salsa, Poultry Aspic with Salsa Garnish, and Fruit Salad with Cilantro Garnish.


EliphasNUIT:
Thanks!

PR #660, opened 2022-09-15
Add Impact Sigil modifier

Zerthox:
This adds Impact Sigil to gear damage modifiers. I also went ahead and replaced skill id literals in Predator's Onslaught with their corresponding constants since I happened to stumble over it while looking at how damage modifiers work in EI.


GW2EIEvtcParser/EIData/DamageModifiers/DamageModifier.cs
```diff
@@ -232,6 +232,7 @@ public IReadOnlyList<AbstractHealthDamageEvent> GetHitDamageEvents(AbstractSingl
             new BuffDamageModifierTarget(NumberOfBoons, "Spellbreaker Rune", "7% on boonless target",  DamageSource.NoPets, 7.0, DamageType.Strike, DamageType.Strike, Source.Gear, ByAbsence, "https://wiki.guildwars2.com/images/1/1a/Superior_Rune_of_the_Spellbreaker.png", DamageModifierMode.All),
             new BuffDamageModifierTarget(Chilled, "Ice Rune", "7% on chilled target",  DamageSource.NoPets, 7.0, DamageType.Strike, DamageType.Strike, Source.Gear, ByPresence, "https://wiki.guildwars2.com/images/7/78/Superior_Rune_of_the_Ice.png", DamageModifierMode.All),
             new BuffDamageModifier(Fury, "Rage Rune", "5% under fury",  DamageSource.NoPets, 5.0, DamageType.Strike, DamageType.Strike, Source.Gear, ByPresence, "https://wiki.guildwars2.com/images/9/9e/Superior_Rune_of_Rage.png", DamageModifierMode.All),
+            new BuffDamageModifierTarget(new long[] { Stun, KnockDown }, "Impact Sigil", "7% on stunned or knocked-down target",  DamageSource.NoPets, 7.0, DamageType.Strike, DamageType.Strike, Source.Gear, ByPresence, "https://wiki.guildwars2.com/images/a/ab/Superior_Sigil_of_Impact.png", DamageModifierMode.All),
```
EliphasNUIT, line 235: Shouldn't that be 10%? Or perhaps the modifier should be renamed to something like "Impact Sigil Efficiency".

Zerthox: 10% would be incorrect since the 3% modifier is **additive** while the 7% against stunned enemies is a separate **multiplicative** modifier. If you feel like it is unclear what the modifier tracks, tthe name could be adjusted.

EliphasNUIT: Oh, I did not know the 7% did not stack additively with the 3%. In that case, everything's fine. Thanks.

PR #671, opened 2022-12-26
Add icon override for the Snowball skill used in the Freezie Strike Mission

Stonos:
**Before:**

image

**After:**

image


EliphasNUIT:
Thanks!

PR #673, opened 2022-12-30
New const and renaming of ids

Linkaaaaa:
A bit of a cleanup of the leftover Ids without consts defined.


EliphasNUIT:
Looks good. Thanks!

PR #678, opened 2023-01-19
Added Bowl of Fancy Potatoe and Leek Soup to tracked buffs

LunusLupus:
Added Bowl of Fancy Potatoe and Leek Soup to tracked buffs


EliphasNUIT:
Thanks!

PR #681, opened 2023-02-11
Refactoring MarkerGUIDs Class & IsCommander Method

Linkaaaaa:
Included the const strings into a HashSet to remove the switch present in the IsCommander method. Additionally, removed the variable _isCommander as it was unnecessary and overall simplified the method.


EliphasNUIT:
Thanks§

PR #684, opened 2023-02-15
Improved ParserHelper and new ParserIcons class

Linkaaaaa:
Removed most switches from the ParserHelper methods, now using TryFind to get the output from dictionaries and moved all the icons-related dictionaries to their own file.


EliphasNUIT:
Thanks!

PR #690, opened 2023-02-21
Harvest Temple Skill IDs

Linkaaaaa:
Added all the named Skill IDs found in logs so far, some have currently temporary names (VoidPool1, VoidPool2) until the correct casting agent is found.
Renamed ScreamOfZhaitan to ScreamOfZhaitanNM as it only shows in normal mode and challenge mode has a different ID,
Reordered some IDs that were missplaced.

~~Magic Discharge might have an additional ID in CM but i have not found it.~~
Correction: Magic Discharge has the same ID in CM.

Cosmic Meteor might have an Ability ID too but i haven't found it, the one in the file is a buff.


EliphasNUIT:
Thanks!

PR #696, opened 2023-03-03
Moved missing Skill IDs

Linkaaaaa:
- Moved missing Skill IDs from some encounter logs to the class
- Changed some of the hard coded IDs to their existing const variables
- Added some new ones.
- Fixed a couple of typos


GW2EIEvtcParser/EncounterLogic/Raids/W6/ConjuredAmalgamate.cs
```diff
@@ -136,7 +136,7 @@ internal override void EIEvtcParse(ulong gw2Build, FightData fightData, AgentDat
             ComputeFightTargets(agentData, combatData, extensions);
             foreach (CombatItem c in combatData)
             {
-                if (c.IsDamage(extensions) && c.SkillID == 52370)
+                if (c.IsDamage(extensions) && c.SkillID == ConjuredSlash2)
```
EliphasNUIT, line 139: ConjuredSlashPlayer would be a clearer name.

PR #697, opened 2023-03-03
Fix mechanic collision with HT green/zhaitan goop

cheahjs:
`Heart Goop and Success Green share the same configuration`


EliphasNUIT:
Thanks!

PR #702, opened 2023-03-09
Added Flux Bomb

Linkaaaaa:
Simple as, flux bomb effect fractal-wide
Called it flux bomb small because i'm not sure yet if it's a different one for the old bigger version.


GW2EIEvtcParser/EncounterLogic/Fractals/FractalLogic.cs
```diff
@@ -144,6 +145,28 @@ internal override void ComputePlayerCombatReplayActors(AbstractPlayer p, ParsedE
                     }
                 }
             }
+            if (fluxBombSmall != null)
```
EliphasNUIT, line 148: That's going to put down the same flux bombs for each player. You can put this in a protected static method and call it in ComputeNPCCombatReplay of each logic under the main target of the fight.

EliphasNUIT:
Thanks!

PR #703, opened 2023-03-11
Update misidentified Void Pool IDs on HT

cheahjs:
Fixed the Void Pool mechanic encompassing the wrong skill IDs.

* `63910` - red AoEs on CM
* `67739` - red AoEs on NM
* `VoidPoolKralkatorrik` - Kralk's AoEs that linger on the arena
* `VoidPoolSooWon` - Soo-Won's AoEs on the arena when they puke black things


EliphasNUIT:
Thanks!

PR #706, opened 2023-03-12
Ensolyss Combat Replay - Mama & Siax Update

Linkaaaaa:
# Generic

- Added Cascade of Torment to MAMA
- Updated the animation of Cascade of Torment for Siax
- Added Nightmare Altar ID
- Added Nightmare Miasma GUIDs for Ensolyss
- Added Nightmare Altar Shockwave GUID
- Renamed some const

# Ensolyss

- Split Upswing mechanic (Ensolyss hits - Hallucinations hits)
- Added phases "Nightmare Altars" between DPS phases
- Added Miasma rings
- Added Arkk's bubble
- Added Tail Lash
- Added Tormenting Blast
- Added Caustic Grasp
- Added Upswing (Ensolyss and Hallucinations)
- Added Lunge (Ensolyss and Hallucinations)
- Added Caustic Explosion

# Notes

- Currently commented Caustic Barrage in ComputeEnvironmentCombatReplayDecorations because the effect is the same for all AoEs but the size varies a lot. I believe the attacks are called "Nightmare Bomb" but I haven't found a way to connect these attacks to a viable agent. If this becomes possible, the barrage decoration needs to be split for each skill cast.
- Currently commented EIEvtcParse which shows the position of the Nightmare Altars, I wanted to use these to apply a red / blue ring during the capture events haven't been able to.
- The combat replay map has a scaling issue, the position of the altars don't match the drawn rings and the miasma is smaller than the size of the map (got the radius from from skilldef).
- The miasma could be displayed without needing the GUID (for old logs) if the cast event gets connected to a valid agent (37151, 37466, 36944)


GW2EIEvtcParser/ParserHelpers/ArcDPSEnums.cs
```diff
@@ -764,6 +764,7 @@ public enum TrashID : int
             // Ensolyss
             NightmareHallucination1 = 16912, // (exploding after jump and charging in last phase)
             NightmareHallucination2 = 17033, // (small adds, last phase)
+            NightmareAltar = 35791,
```
EliphasNUIT, line 767: That looks like a GadgetID, unlikely to be reliable as they tend to be volatile.

EliphasNUIT:
Thanks!

PR #708, opened 2023-03-13
Restored arkk's shield functionality

Linkaaaaa:
The changes made last night don't allow for more than shield to appear in the combat replay. I've also restored the functionality of bool shield15_0Added because it's needed in case the 15% bubble is skipped and it's used to prevent multiple decorations from being added (the 15% keeps reapplying the shield effect every few seconds).


EliphasNUIT:
Thanks!

PR #715, opened 2023-04-07
Reward classes

Linkaaaaa:
- Created classes RewardTypes and RewardIDs
- Moved old and current raid rewards and eod reward to RewardTypes
- Moved a reward ID for ibs strike to RewardIDs and added IDs for the other strikes

Bonus:
- Streamlined Energy Threshold since it's the same id for Dhuum and Eater
- Added the build number of the raid rewards change to GW2Builds enum
- Added IDs for racial summons and the deluxe edition of the core game


EliphasNUIT:
Thanks!

PR #716, opened 2023-04-09
Fix MAMA log times

Zerthox:
Players entering combat with Knights or an invisible hitbox before attacking MAMA results in incorrect log times compared to triggering a log by the first player entering combat when attacking the boss. This happens fairly often, since a player placing a damaging AoE on MAMA during the spawn process likely hits the invisible hitbox. The time difference in logs is caused by MAMA's spawning Invulnerability being recorded, which is not present in regular logs since they start at the first attack hitting MAMA without Invulnerability. EI then takes the Invulnerability running out as the starting point instead of the first hit on MAMA.

This PR attempts to reduce the variation by using the time of the first relevant event regarding MAMA. It should end up producing the exact same start times for logs triggered by attacking the boss and result in later & more comparable start times for logs triggered by an early player combat enter.


Zerthox:
Additional note: this isn't perfect yet since some buffs (e.g. Revealed) can be applied during the Invulnerability phase. Maybe the first buff apply should be limited to specific buffs?

Zerthox:
Also missing the inverse combat trigger - MAMA hitting a player.

Under what circumstances would combat enter for the boss be missing from the log? It is complicated to filter exactly the events that would cause a combat enter for MAMA.

EliphasNUIT:
Wouldn't it be simpler to adapt the method in Fractal.cs? I don't think it is okay to base the beginning of an encounter on player behavior as it can be easily exploited. EnterCombatEvent will be missing on logs where the logging starts after MAMA is already in combat, should happen much less nowadays with the new arc log starts but will be much more frequent on logs from 2022 and before.
In such logs, in your PR, I can make MAMA run after me while not hitting it or getting hit and that would artificially increase the start of the encounter.
I believe it's better to update GetFightOffsetByFirstInvulFilter to select the proper time between CombatEnter.Time and InvulLost.Time (that would also benefit fractals as a whole) as player not doing anything after those two would be on them, not on the fight.

Zerthox:
> In such logs, in your PR, I can make MAMA run after me while not hitting it or getting hit and that would artificially increase the start of the encounter.

That is precisely what you should be able to do. Contrary to all other Fractal CM bosses, the MAMA encounter does not automatically put players in combat. Instead players enter combat and thus logs start when hitting or getting hit by MAMA.

I don't think with current ArcDPS logging there is a way to start a MAMA log after MAMA already entered combat? And I would be fine to only apply the"fix" to logs recorded after the ArcDPS logging change.

GW2EIEvtcParser/ParsedData/FightData.cs
```diff
@@ -13,6 +13,7 @@ public class FightData
         private List<PhaseData> _phases = new List<PhaseData>();
         private List<PhaseData> _nonDummyPhases = new List<PhaseData>();
         public int TriggerID { get; }
+        public int EvtcVersion { get; }
```
EliphasNUIT, line 16: That attribute is already on LogData, I don't think it is a good idea to duplicate it here. FightData is also not the best place to store evtc related meta data.
You can simply modify the signature of GetFightOffset and give evtcVersion to it, the method is called in EvtcParser.cs where the value also exists.

PR #719, opened 2023-04-17
Fix Mushroom King's Blessing detection

Zerthox:
This updates the invisible buff id used to determinte casts of [Mushroom King's Blessing](https://wiki.guildwars2.com/wiki/Mushroom_King%27s_Blessing) in the Special Forces Training Area. Out of the 4 invisible buffs applied it uses the one that seems to be responsible for granting the skill recharge reduction.

It also adds name & icon overrides for the skill.

Edit: ~~for clarification, the current id is not working as its different from all 4 buffs applied.~~ Because skill id and buff id were mixed up.


Zerthox:
It seems like there was a mix up with the buff & skill id here previously. I'll adjust it to use the proper skill id as skill id.

EliphasNUIT:
Thanks!

PR #721, opened 2023-04-18
Target name improvements

Zerthox:
Improvements to target names, making sure their numbers makes sense or using location suffixes (N for north, E for east etc.).

- Siax: Adjusts the target name to be "Siax the Corrupted" rather than "Nightmare Oratuss"
- Siax: Adds location suffix to "Echo of the Unclean" split targets
- Skorvald: Adjusts the Anomaly numbers to be in the order you (usually) kill them
- Artsariiv: Adds location suffix to Clone split targets

### DIFF
--- a/GW2EIEvtcParser/EncounterLogic/Fractals/ShatteredObservatory/Skorvald.cs
+++ b/GW2EIEvtcParser/EncounterLogic/Fractals/ShatteredObservatory/Skorvald.cs
@@ -98,15 +98,21 @@ namespace GW2EIEvtcParser.EncounterLogic
                 throw new MissingKeyActorsException("Skorvald not found");
             }
             skorvald.OverrideName("Skorvald");
-            int count = 0;
             foreach (NPC target in _targets)
             {
-                if (target.IsSpecies(ArcDPSEnums.TrashID.FluxAnomaly1) ||
-                    target.IsSpecies(ArcDPSEnums.TrashID.FluxAnomaly2) ||
-                    target.IsSpecies(ArcDPSEnums.TrashID.FluxAnomaly3) ||
-                    target.IsSpecies(ArcDPSEnums.TrashID.FluxAnomaly4))
-                {
-                    target.OverrideName(target.Character + " " + (++count));
+                switch (target.ID) {
+                    case (int) ArcDPSEnums.TrashID.FluxAnomaly1:
+                        target.OverrideName(target.Character + " 1");
+                        break;
+                    case (int) ArcDPSEnums.TrashID.FluxAnomaly2:
+                        target.OverrideName(target.Character + " 2");
+                        break;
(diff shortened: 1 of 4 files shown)


Zerthox:
Name collisions in the same split phase should not be possible with this.

Collisions are possible across different split phases, e.g. there is a separate "Anomaly 1" for both Skorvald split phases or a separate "Clone Artsariiv NW" for both Artsariiv split phases. Is this problematic? If so, I guess one option would be to have Skorvald Anomalies go from 1-8 again and Artsariiv clones could be "NW 1" and "NW 2". Siax target names should be unique across both split phases anyway.

Zerthox:
Names should be unique across phases now.

I realized I accidentally removed the "Clone" part from the Artsariiv clone targets, do you want to add it back and make it "Clone Artsariiv NW 1" or is "Artsariiv NW 1" enough?

EliphasNUIT:
I'd say the clone is not necessary now that the cardinals are in the names.
To answer your first question, there are parts in the HTML where targets are enumerated so having multiple ones with the same name could lead to confusions.

GW2EIEvtcParser/EncounterLogic/Fractals/ShatteredObservatory/Skorvald.cs
```diff
@@ -73,12 +73,32 @@ internal override List<PhaseData> GetPhases(ParsedEvtcLog log, bool requirePhase
...
-                        (int)ArcDPSEnums.TrashID.FluxAnomaly2,
                         (int)ArcDPSEnums.TrashID.FluxAnomaly1,
+                        (int)ArcDPSEnums.TrashID.FluxAnomaly2,
+                        (int)ArcDPSEnums.TrashID.FluxAnomaly3,
+                        (int)ArcDPSEnums.TrashID.FluxAnomaly4,
                     };
                     AddTargetsToPhaseAndFit(phase, ids, log);
+
+                    // add anomaly numbers
+                    int offset = 4 * (i / 2 - 1);
```
EliphasNUIT, line 84: Having the overrides be in GetPhases is a little bit problematic as that means that as long as GetPhases method hasn't been called, the name of the targets would be wrong.
Something like Dictionary<TrashID, int> { {FluxAnomaly1, 1} ... {FluxAnomaly4, 1}} where the count is fetched and incremented in EIEvtcParse would be better.
Same for the Artsariiv clones, a dictionary based on CloneLocations and have your utility method return the matched <string, float, float > maybe?

Zerthox: I see. For Skorvald that could work. For Artsariiv the N, E, S, W clones only appear in the second split. I would still prefer to name the first occurrence of those "Artsariiv N 2" after the phase they occur in, if that's fine?

EliphasNUIT:
Thanks!

PR #724, opened 2023-04-24
Add Skorvald normal mode Anomalies

Zerthox:
Adds ids for Flux Anomalies in normal mode Skorvald. This allows normal mode Skorvald to also display split phases.


EliphasNUIT:
Thanks

PR #727, opened 2023-04-27
Add old gyros

Zerthox:
Adds ids for the old gyro minions.

### DIFF
--- a/GW2EIEvtcParser/ParserHelpers/ArcDPSEnums.cs
+++ b/GW2EIEvtcParser/ParserHelpers/ArcDPSEnums.cs
@@ -1206,7 +1206,9 @@ namespace GW2EIEvtcParser
             FireElemental = 6524,
             IceElemental = 6525,
             //
+            SneakGyro = 15012,
             ShredderGyro = 15046,
+            BulwarkGyro = 15134,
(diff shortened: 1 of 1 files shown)


GW2EIEvtcParser/ParserHelpers/ArcDPSEnums.cs
```diff
@@ -1206,6 +1206,13 @@ public enum MinionID : int
             FireElemental = 6524,
             IceElemental = 6525,
             //
+            SneakGyro = 15012,
```
EliphasNUIT, line 1209: Any intent to use those somewhere down the line? Or just putting them down for reference?

Linkaaaaa: It's just for historical value, I will add the icons to ParserIcons later on. I'm waiting on a friend to make them so they can show up in the combat replay (does that even exist for 2018-2019 logs?)

EliphasNUIT: 2018 is very unlikely as movement data for non players/non boss ids were introduced much later, 2019 should be fine~ish depending on the date. Would need to be checked.

PR #729, opened 2023-04-29
Instant cast detections

Zerthox:
- Necromancer Shadow Fiend Haunt detection by buff applied to the player
- Necromancer Bone Fiend Rigor Mortis detection by buff applied to the minion
- Necromancer Flesh Wurm Necrotic Traversal detection by effect around dying minion
- Helpers for common spec checking patterns on effect-based detection
- Adjusted skill id names to follow the naming convention


Zerthox:
It seems that whenever the player commands a minion to perform a skill, the game applies a generic buff (59536) to the minion. For Necromancers & Elementalists this is applied by the player. Ranger pets applied it to themselves. The buff is **not** used by the Mechanist Mech.

The buff appears to not be pov only. I've added a new instant cast finder to make use of this buff.

GW2EIEvtcParser/ParserHelpers/EffectGUIDs.cs
```diff
@@ -7,6 +7,8 @@ namespace GW2EIEvtcParser
 {
     public static class EffectGUIDs
     {
+        // Generic
+        public const string Generic600UnitRadius = "9C7C1B2379CCDD4990001A38030E4495";
```
EliphasNUIT, line 11: What is that?

Zerthox: Effect displaying a radius indicator of 600 units. It's used for several skills and pulsing effects. Thought it might be useful to note down for future reference, so you can recognize and ignore it when looking into effects for a skill.

GW2EIEvtcParser/EIData/InstantCastFinders/MinionCastCastFinder.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using System.Linq;
+using GW2EIEvtcParser.ParsedData;
+
+namespace GW2EIEvtcParser.EIData
+{
+    internal class MinionCastCastFinder : CheckedCastFinder<AnimatedCastEvent>
```
EliphasNUIT, line 7: Regarding this class, are we sure that a minion casting something necessarily matches the moment the command is given? I'm guessing you want to use this for the Jade Mech?

Zerthox: No, it will not necessarily match the moment the command is given. The idea was to use this potentially for some of the Mech & Untamed Pet skills. I believe for those it should be good enough, since they are very close to actually pressing the skill for the Mech/Pet and can even queue up.

GW2EIEvtcParser/EIData/ProfHelpers/Engineer/MechanistHelper.cs
```diff
@@ -39,7 +51,7 @@ internal static class MechanistHelper
             new Buff("Force Signet", ForceSignet, Source.Mechanist, BuffClassification.Other, BuffImages.ForceSignet),
             new Buff("Shift Signet", ShiftSignetEffect, Source.Mechanist, BuffClassification.Other, BuffImages.ShiftSignet),
             new Buff("Superconducting Signet", SuperconductingSignet, Source.Mechanist, BuffClassification.Other, BuffImages.SuperconductingSignet),
-            new Buff("Overclock Signet", OverclockSignet, Source.Mechanist, BuffClassification.Other, BuffImages.OverclockSignet),
+            new Buff("Overclock Signet", OverclockSignetSkill, Source.Mechanist, BuffClassification.Other, BuffImages.OverclockSignet),
```
EliphasNUIT, line 54: Wrong ID used here

Zerthox:
Regarding the Overclock Signet, do you think instead of marking it as not accurate it would be better to use Jade Buster Cannon as the instant cast skill?

EliphasNUIT:
Well, that's not a skill that is actually in the mechanist's bar so I'd rather keep the signet.

GW2EIEvtcParser/EIData/ProfHelpers/Mesmer/MesmerHelper.cs
```diff
@@ -265,10 +259,16 @@ private static bool IsClone(long id)
...
         };
+
         internal static bool IsKnownMinionID(long id)
         {
             return NonCloneMinions.Contains(id) || IsClone(id);
         }
 
+        internal static bool spawnedClone(AgentData agentData, MinionID clone, AgentItem master, long time)
+        {
+            return agentData._allAgentsList
```
EliphasNUIT, line 270: This won't compile in RELEASE, use agentData.GetNPCsByID(clone) instead

GW2EIEvtcParser/EIData/ProfHelpers/Mesmer/MirageHelper.cs
```diff
@@ -14,7 +14,8 @@ internal static class MirageHelper
 
         internal static readonly List<InstantCastFinder> InstantCastFinder = new List<InstantCastFinder>()
         {
-            new DamageCastFinder(Jaunt, Jaunt), // Jaunt
+            new DamageCastFinder(Jaunt, Jaunt).UsingEnable(combatData => !combatData.HasEffectData),
+            new EffectCastFinderByDst(Jaunt, EffectGUIDs.MirageJaunt).UsingDstSpecChecker(Spec.Mirage),
```
EliphasNUIT, line 18: Are you sure about that? I remember having to remove Jaunt effect finder as the 3 effect IDs I had identified were ultimately colliding with something else.

Zerthox: I didn't manage to find a conflict with the skills I tested manually (Staff 2, Axe 3, Blink, Illusionary Ambush). It's possible it collides with something I did not test. Do you have any idea what it could've collided with?

EliphasNUIT: Unfortunately I don't remember as that was nearly a year ago, however Jaunt is in my "checked" list and marked as all IDs colliding. The tested golem log contains the effect IDs you have added. It may be worth double checking however.

Zerthox:
The buff detection for rise seems to collide with [Summon Madness](https://wiki.guildwars2.com/wiki/Summon_Madness) from Lich Form. Will see if it can be improved.

EliphasNUIT:
Thanks!

PR #732, opened 2023-05-02
Categorize Thief summons

Zerthox:
This categorizes some of the Thief summons. They seem to use the same above and underwater. Not sure why there is so many of them. Possibly RNG, possibly based on character gender & race? Will have to investigate some more.


Zerthox:
They indeed seem to be different depending on character race (gender seems to have no effect?).

EliphasNUIT:
Thanks!

PR #735, opened 2023-05-03
More instant cast detections

Zerthox:
- Scourge Shade skill detection by damage/buff
- Necromancer Plague Signet cast detection by effect
- Thief Preparation skill detection by effect
- Thief Signet skill detection by effect
- Elementalist Signet of Air detection by effect (instead of damage)
- Mechanist Discharge Array detection by buff (instead of inaccurate effect)
- Fix Mechanist Crisis Zone effect
- Mechanist Overclock Signet detection by effect (instead of inaccurate minion cast)
- Guardian Meditation, Shout & Signet skill detection by effect
- Mesmer Signet of Midnight detection by effect
- Ranger Signet skill detection by effect
- Fix Ranger Signet buff names
- Warrior Signet skill detection by effect
- Some additional effect GUIDs
- Some cleanup regarding skill names & a note regarding POV only skills/buffs
- Helper functions

Is `BuffLossCastFinder` potentially triggered when losing buffs related to an agent down/dead event? If so, maybe a way to filter out those triggers make sense.


EliphasNUIT:
Depending on the buff, down and dead events may trigger a buff loss, yes.

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/NecromancerHelper.cs
```diff
@@ -28,6 +28,7 @@ internal static class NecromancerHelper
             new BuffGainCastFinder(SpectralArmorSkill, SpectralArmorEffect).WithBuilds(GW2Builds.December2018Balance),
             new BuffGainCastFinder(SpectralWalkSkill, SpectralWalkEffectOld).WithBuilds(GW2Builds.StartOfLife, GW2Builds.December2018Balance),
             new BuffGainCastFinder(SpectralWalkSkill, SpectralWalkEffect).WithBuilds(GW2Builds.December2018Balance),
+            new BuffLossCastFinder(PlagueSignetSkill, PlagueSignetEffect),
```
EliphasNUIT, line 31: This will trigger on downs/deads. It is better to check if an effect exists that matches the cast of the skill. Checking the timestamps of down and dead events can be a bit shaky, the timing threshold will need to be definitively higher than ServerDelayConstant.

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
```diff
@@ -10,17 +10,18 @@ namespace GW2EIEvtcParser.EIData
 {
     internal static class ScourgeHelper
     {
-
         internal static readonly List<InstantCastFinder> InstantCastFinder = new List<InstantCastFinder>()
         {
             // Trail of Anguish ? Unique effect?
             new EffectCastFinder(TrailOfAnguish, EffectGUIDs.ScourgeTrailOfAnguish).UsingSrcSpecChecker(Spec.Scourge).UsingICD(6100),
```
EliphasNUIT, line 16: If you found a non PoV only buff for trails of anguish, it is better to use it for the InstantCast detection, buff based solutions are always the best.

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/NecromancerHelper.cs
Zerthox: This would also apply to the current Signet of Midnight detection for Mesmer then. Will have to check whether the unique effects are POV only.

EliphasNUIT: That would require the Mesmer going down/dying at the same time as applying 2 seconds of Hide in Shadows to themself, which sounds unlikely if not impossible.

Zerthox: Turns out the Signet effect for Necromancer is **not** POV-only. It should be possible to detect other instant cast Signets using their respective effects. What do you think about adding effect-based detection and keeping the current Signet of Midnight detection for logs without effects?

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
Zerthox: The 42311 one seems to be the only non-POV-only out of them, but for some reason it has twice the duration (12s vs 6s). I'll include it for now.

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/NecromancerHelper.cs
EliphasNUIT: They can be effect based, yes. Except the ones giving a unique buff on signet activation (signet of stone, assassin's signet,...).

GW2EIEvtcParser/EIData/ProfHelpers/Guardian/GuardianHelper.cs
```diff
@@ -13,20 +16,43 @@ internal static class GuardianHelper
...
+            new DamageCastFinder(JudgesIntervention, JudgesIntervention).UsingDisableWithEffectData(),
+            new EffectCastFinderByDst(JudgesIntervention, EffectGUIDs.GuardianJudgesIntervention).UsingDstBaseSpecChecker(Spec.Guardian),
+            new EffectCastFinderByDst(ContemplationOfPurity, EffectGUIDs.GuardianContemplationOfPurity1).UsingDstBaseSpecChecker(Spec.Guardian),
+            new DamageCastFinder(SmiteCondition, SmiteCondition),
+            new DamageCastFinder(LesserSmiteCondition, LesserSmiteCondition),
+            
+            // Shouts
+            new EffectCastFinderByDst(SaveYourselves, EffectGUIDs.GuardianSaveYourselves).UsingDstBaseSpecChecker(Spec.Guardian),
+            // distinguish by boons, check duration/stacks to counteract pure of voice
+            new EffectCastFinderByDst(Advance, EffectGUIDs.GuardianShout)
```
EliphasNUIT, line 36: This should be specifically flagged as NotAccurate, in case effect events are not NotAccurate by default in the futur.

GW2EIEvtcParser/EIData/ProfHelpers/Guardian/GuardianHelper.cs
```diff
@@ -13,20 +16,43 @@ internal static class GuardianHelper
...
+                .UsingDstBaseSpecChecker(Spec.Guardian)
+                .UsingChecker((evt, combatData, agentData, skillData) =>
+                {
+                    return FindRelatedEvents(combatData.GetBuffData(Aegis).OfType<BuffApplyEvent>(), evt.Time)
+                        .Any(apply => apply.By == evt.Dst && apply.To == evt.Dst && apply.AppliedDuration >= 20000 && apply.AppliedDuration <= 40000);
+                }),
+            new EffectCastFinderByDst(StandYourGround, EffectGUIDs.GuardianShout)
```
EliphasNUIT, line 43: Same here, NotAccurate.

Zerthox:
What should the naming for secondary skills that are responsible for hits from a different skill be? I've suffixed them with "Hit" (and "Multi Hit") now, but would "Strike" be preferred instead?

EliphasNUIT:
Extra hit maybe?
It seems you have conflicts coming from your other PR, could you solve them please?

Zerthox:
The skill overrides might need some cleanup/double checking.

EliphasNUIT:
Thanks!

PR #736, opened 2023-05-03
Fix Dhuum phase detection bug

Zerthox:
This broken log was posted on the Wingman Discord: https://gw2wingman.nevermindcreations.de/log/KO0o-20230502-204904_dhuum.
Turns out the phase detection failed since in the log Dhuum casts Weapon Draw 2ms into the encounter.


EliphasNUIT:
Thanks

PR #747, opened 2023-05-06
Fix Harvest Temple Skill IDs

Linkaaaaa:
The IDs of the red Void Pools in normal mode and Soo Won's Void Pools were inverted


EliphasNUIT:
Thanks!

PR #749, opened 2023-05-08
Fix instant cast detections

Zerthox:
This fixes effect collisions with Thousand Needles (e.g. Deadeye Shadow Flare). It also adds a way to check for secondary effects. The new cast detection for Thousand Needles happens ~280ms after the related buff is lost (and the previously used effect happens), right as the first hit occurs.

In the future we might want to detect it by buff loss and filter out buff removes too close to dead events.


GW2EIEvtcParser/EIData/ProfHelpers/Thief/ThiefHelper.cs
```diff
@@ -24,7 +26,10 @@ internal static class ThiefHelper
             new BuffGiveCastFinder(SoulStoneVenomSkill,SoulStoneVenomEffect),
             new BuffGiveCastFinder(SpiderVenomSkill,SpiderVenomEffect).UsingChecker((evt, combatData, agentData, skillData) => evt.To != evt.By || Math.Abs(evt.AppliedDuration - 24000) < ServerDelayConstant).UsingNotAccurate(true), // same id as leeching venom trait?
             new EffectCastFinder(Pitfall, EffectGUIDs.ThiefPitfallAoE).UsingSrcBaseSpecChecker(Spec.Thief),
-            new EffectCastFinder(ThousandNeedles, EffectGUIDs.ThiefThousandNeedlesAoE1).UsingSrcBaseSpecChecker(Spec.Thief),
+            new BuffLossCastFinder(ThousandNeedlesArmedEffect, ThousandNeedlesArmedEffect),
+            new EffectCastFinder(ThousandNeedles, EffectGUIDs.ThiefThousandNeedlesAoE1)
```
EliphasNUIT, line 30: I don't see anything solving the 280 ms delay issue.

Zerthox: I can add an offset, but the delay can be less than 280ms. I think having the instant cast show up in the log before the actual cast happened is worse than having it show up around the first hit?

Zerthox:
Also fixed false positives for Judges Intervention, Merciful Intervention, Crisis Zone (maybe?). And made Save Yourselves and You Are All Weaklings and Infiltrator's Signet more robust by intersecting all effects.

GW2EIEvtcParser/ParserHelpers/SkillIDs.cs
```diff
@@ -221,7 +221,7 @@ public static class SkillIDs
         public const long SignetOfResolve = 9220;
         public const long BindingBladeSelf = 9225;
         public const long MercifulInterventionTargetEffect = 9231;
-        public const long MercifulInterventionSelfEffect = 9235;
+        public const long MercifulInterventionSelfEffect = 9235; // also used by judges intervention
```
EliphasNUIT, line 224: Same, should be renamed.

GW2EIEvtcParser/EIData/ProfHelpers/Thief/ThiefHelper.cs
EliphasNUIT: I don't understand your point, how the instant cast detection could trigger before the actual cast happens? Yes, it is not correct to show the cast after or before it happened, if the timing of the cast is wrong and can only be wrong with given detection than it should be removed.
I don't see how this related to this situation however? You said the colliding effect was the trigger with the other 2 delayed by around 280ms but I see now you are using a buff loss while checking those two effects with a delay, are the buff loss and colliding effect equivalent? If yes, and the buff loss corresponds to the moment the cast fires, I don't see any problem with the proposed implementation.

Zerthox:
Crisis Zone might still need more looking into, we discovered one of the effects can happen without the other. Either one effect is reused or the other does not happen every time.

Zerthox:
The effect is also used delayed after the Mech spawns with Crash Down. I swapped the effects so that the reused effect is a secondary check.

EliphasNUIT:
This is getting a little bit confusing for me. Could you summarize the status for the following please (latest release combined with what is in this PR)?

- Which instant casts still currently collide with something else?
- Which instant casts still trigger less than expected?
- Which instant casts detection timings are not correct ?

Thanks.

Zerthox:
This PR fixes collisions for:
1. Thief **Thousand Needles** false positive triggered by Shadow Flare (effect reuse). Fixed by detecting cast via buff instead and checking for related effects (marked as not accurate due to effect usage).
2. Guardian **Merciful Intervention** false positive triggered by Judges Intervention (buff reuse). Fixed by checking for related effects like above.
3. Guardian **Judges Intervention** false positive triggered by Symbol of Blades (effect reuse). Fixed via changing to copy of above but with different effect to distinguish.
4. Mechanist **Crisis Zone** false positive triggered by Crash Down (effect reuse). Fixed by checking secondary effect.
5. Mesmer **Blink** false positive & negative. Simply a typo in the code.

As of now I am not aware of any other known false positives or false negatives. The timings should continue to be correct now with the Buff + Effect hybrid detections.

Zerthox:
Everything ready.

EliphasNUIT:
Thanks!

Issue #740, opened 2023-05-04
[Enhancement] Add mechanic count for players lifted up at Qadim the Peerless


Linkaaaaa:
Add a mechanic counter for the players that got lifted up in the air at Qadim the Peerless.

Linkaaaaa:
The player lifted up in the air receives these 2 visible buffs:
Rebellious Power: 56434 [&BnLcAAA=] 
Charged Soul: 56011 [&BsvaAAA=]

Linkaaaaa:
Implemented in #750

PR #750, opened 2023-05-09
General updates

Linkaaaaa:
- Fixed the buff icons for Rigom and Guldhem Fixate
- Fixed the naming scheme of Mursaat Overseer special action key buffs and added the special action keys skill ids for completition
- Added two new buffs to Qadim the Peerless CM which appear when a player gets lifted up in the air ( #740 )
- Added a mechanic to Qadim the Peerless CM which counts how many times a player has been lifted up ( #740 )
- Added two mechanics to Qadim the Peerless which count the special action key activation and deactivation cast times
- Added a buff to Kaineng Overlook strike
- Corrected the naming of the Basilisk Venom buff and added a renaming with icon override for its stun effect
- Added names to the Qadim the Peerless special action keys cast time, which have a different skill ID than the skill itself
- Added icon override for Sabir's and all three Qadim the Peerless' special action keys


EliphasNUIT:
Thanks!

PR #753, opened 2023-05-13
More instant cast detections

Zerthox:
- Harbinger Cascading Corruption, Deathly Haste, Doom Approaches detection by effect


GW2EIEvtcParser/EIData/ProfHelpers/Engineer/MechanistHelper.cs
```diff
@@ -20,6 +21,19 @@ internal static class MechanistHelper
...
             // Mech
+            new DamageCastFinder(CrashDown, CrashDown).UsingDisableWithEffectData(),
+            new MinionSpawnCastFinder(CrashDown, (int)MinionID.JadeMech)
+                .UsingChecker((spawn, combatData, agentData, skillData) =>
+                {
+                    // TODO: what if shift signet ports the minion away from effect before arc polls the position? so far unable to produce
+                    Point3D pos = combatData.GetMovementData(spawn.Src).OfType<PositionEvent>().FirstOrDefault(evt => evt.Time + ServerDelayConstant >= spawn.Time)?.ToPoint();
+                    if (pos != null && combatData.TryGetEffectEventsByGUID(EffectGUIDs.MechanistCrashDownImpact, out IReadOnlyList<EffectEvent> effects))
+                    {
+                        return FindRelatedEvents(effects, spawn.Time + 800).Any(effect => pos.Distance2DToPoint(effect.Position) < 10.0f);
```
EliphasNUIT, line 32: Should be changed to Count(...) == 1 instead of Any(...) for the very rare situations where two crash downs happen  at the same time on the same place.
However, wasn't Crash Down an animated skill for the Jade Mech?

Zerthox: I think if two Mech Crash Downs at the same place `Any(...)` should work fine? `Count(...) == 1` would mean neither of the two Crash Down casts get detected in that case. The scenario where `Any(...)` does not work correctly is when Mech A crashes down at the location and Mech B spawns there without Crash Down. Not sure if that is a realistic thing to happen? To be honest I don't even know under what circumstances you could get a spawn event for a Mech without Crash Down.

Crash Down clearly plays an animation ingame, but it seems Arc does not detect it...? Looking at logs the first cast from the Mechs spawning via Crash Down is usually some weapon drawing ~3s later.

GW2EIEvtcParser/EIData/ProfHelpers/Ranger/DruidHelper.cs
```diff
@@ -14,6 +14,7 @@ internal static class DruidHelper
         {
             new BuffGainCastFinder(EnterCelestialAvatar, CelestialAvatar).UsingBeforeWeaponSwap(true),
             new BuffLossCastFinder(ExitCelestialAvatar, CelestialAvatar).UsingBeforeWeaponSwap(true),
+            new BuffGainCastFinder(AncestralGraceSkill, AncestralGraceEffect),
```
EliphasNUIT, line 17: That's not an instant cast but a movement skill, It should follow a similar pattern to ComputeFlowingResolveCastEvents in WillbenderHelper.cs.

GW2EIEvtcParser/EIData/ProfHelpers/Engineer/HolosmithHelper.cs
```diff
@@ -45,6 +45,11 @@ public static bool IsPhotonForgeTransform(long id)
             new Buff("Afterburner", Afterburner, Source.Holosmith, BuffStackType.StackingConditionalLoss, 25, BuffClassification.Other, BuffImages.SolarFocusingLens),
             new Buff("Heat Therapy", HeatTherapy, Source.Holosmith, BuffStackType.Stacking, 10, BuffClassification.Other, BuffImages.HeatTherapy),
             new Buff("Overheat", OverheatEffect, Source.Holosmith, BuffClassification.Other, BuffImages.Overheat),
+
+            // heat buffs only present in forge
+            new Buff("0-50 Heat", PhotonForgeAbove0Heat, Source.Holosmith, BuffClassification.Unknown, BuffImages.Heat),
```
EliphasNUIT, line 50: Why Unknown classification?

GW2EIEvtcParser/EIData/ProfHelpers/Ranger/DruidHelper.cs
```diff
@@ -19,6 +20,19 @@ internal static class DruidHelper
...
 
+        public static IReadOnlyList<AnimatedCastEvent> ComputeAncestralGraceCastEvents(Player druid, CombatData combatData, SkillData skillData, AgentData agentData)
+        {
+            var res = new List<AnimatedCastEvent>();
+            SkillItem skill = skillData.Get(AncestralGraceSkill);
+            var applies = combatData.GetBuffData(AncestralGraceEffect).OfType<BuffApplyEvent>().Where(x => x.To == druid.AgentItem).ToList();
+            var removals = combatData.GetBuffData(AncestralGraceEffect).OfType<BuffRemoveAllEvent>().Where(x => x.To == druid.AgentItem).ToList();
+            for (int i = 0; i < applies.Count && i < removals.Count; i++)
+            {
+                res.Add(new AnimatedCastEvent(druid.AgentItem, skill, applies[i].Time, applies[i].AppliedDuration - removals[i].RemovedDuration));
```
EliphasNUIT, line 31: Replace "applies[i].AppliedDuration - removals[i].RemovedDuration" with "removals[i].Time - applies[i].Time" to be more robust.

GW2EIEvtcParser/EIData/ProfHelpers/Engineer/HolosmithHelper.cs
Zerthox: Added icons & changed the names to e.g. "0-50 Heat (Photon Forge)". You think that's clear enough to keep displaying the percentage in the buffs tab?

EliphasNUIT: Looks good to me

Zerthox:
Linka and I think it would be a good idea to change the naming convention for buff/debuff effect skill ids from "effect" to "buff". The rest of the codebase uses "buff" for buff effects (the buff event classes, buff instant finders, methods etc.) and uses "effect" for visual effects. Since this PR already includes a large cleanup for the skill name/icon overrides I think it might be a suitable place to include that change?

Other than that I don't have anything additional for this PR at the moment, so I will probably mark it as ready for merge soon.

EliphasNUIT:
I'd rather have that in another PR than here, I'd rather not deal with that in the review of this PR.

Zerthox:
Do you want me to move the icon/name cleanup to the other PR as well then?

GW2EIEvtcParser/EIData/InstantCastFinders/CastFinderHelpers.cs
```diff
@@ -50,6 +50,12 @@ internal static bool HasGainedBuff(CombatData combatData, long buffID, AgentItem
                 .Any(apply => apply.To == agent && apply.CreditedBy == source && Math.Abs(apply.AppliedDuration - appliedDuration) < epsilon);
         }
 
+        internal static bool HasLostBuff(CombatData combatData, long buffID, AgentItem agent, long time, long epsilon = ServerDelayConstant)
+        {
+            return FindRelatedEvents(combatData.GetBuffData(buffID).OfType<AbstractBuffRemoveEvent>(), time, epsilon)
```
EliphasNUIT, line 55: That would also trigger on BuffRemoveSingle events, is that the intention behind the method, as the buff wouldn't be completely lost but only one stack removed? Otherwise would be better to limit to BuffRemoveAll events. We can add a method to specifically track one stack getting removed later if needed.

Zerthox: It was intentional since I don't know how many stacks get removed by Swap when you would somehow manage to get more than 1 stack of the Illusionary Leap buff. No idea if there is even a way to achieve that. I think it makes sense to rename this to something like `HasLostBuffStack` and add another one for losing last/all.

EliphasNUIT:
Thanks!

PR #758, opened 2023-05-22
Whisper of Jormag Combat Replay

LunusLupus:
- Added Combat replay map for Whisper of Jormag encounter
- Repositioned Combat replay map for Whisper of Jormag encounter


EliphasNUIT:
Thanks!

PR #761, opened 2023-05-27
Portal decorations in combat replay

Zerthox:
Portal decorations for:
- Mesmer Portal (inactive & active)
- Thief Shadow Portal
- Scourge Sand Swell
- White Mantle/Watchwork Portal


GW2EIEvtcParser/ParsedData/CombatData.cs
```diff
@@ -998,6 +998,18 @@ public IReadOnlyList<EffectEvent> GetEffectEventsByTrackingID(long trackingID)
             return new List<EffectEvent>();
         }
 
+        public bool TryGetEffectEndByTrackingId(long trackingID, long time, out long end)
+        {
+            end = 0;
+            EffectEvent endEvent = GetEffectEventsByTrackingID(trackingID).FirstOrDefault(effect => effect.EffectID == 0 && effect.Time >= time);
```
EliphasNUIT, line 1004: I haven't been using it at this point, but it'd be cleaner to have a property IsEnd => EffectID == 0 in EffectEvent.cs and use that instead for better readability

Zerthox:
Should the `GetEffectsForPlayer` and `GetGroupedEffectsForPlayer` helpers be moved somewhere else? Currently they are in `ProfHelper`.

EliphasNUIT:
They can be moved to CombatData as a Public API and remove "Player" notion from it to work directly with an AgentItem like other GetXXXEvents.

EliphasNUIT:
Like GetEffectEventsBySrcWithGUID or something

GW2EIEvtcParser/EIData/ProfHelpers/Mesmer/MesmerHelper.cs
```diff
@@ -323,5 +323,39 @@ internal static bool IsKnownMinionID(long id)
...
+        
+        internal static void ComputeProfessionCombatReplayActors(AbstractPlayer player, ParsedEvtcLog log, CombatReplay replay)
+        {
+            // TODO: tracking ids seem broken in arc? using buff remove with hardcoded fallback for now
+            foreach (EffectEvent effect in ProfHelper.GetEffectsForPlayer(log.CombatData, player, EffectGUIDs.MesmerPortalInactive))
+            {
+                int start = (int)effect.Time;
+                var remove = log.CombatData.GetBuffData(PortalWeaving).OfType<AbstractBuffRemoveEvent>().FirstOrDefault(x => x.Time >= start);
+                int end = (int?)remove?.Time ?? start + 60000;
+                replay.Decorations.Add(new IconDecoration("https://wiki.guildwars2.com/images/8/81/Portal_Entre.png", 128, 0.5f, effect.Src.ID, (start, end), new PositionConnector(effect.Position)));
```
EliphasNUIT, line 335: Src.UniqueID, ID is the SpeciesID

EliphasNUIT: You should simply give an AgentItem to the Constructor and have the constructor put the proper value on the attribute.

GW2EIBuilders/Resources/JS/CR-JS/decorations.js
```diff
@@ -649,4 +649,29 @@ class MovingPlatformDrawable extends BackgroundDrawable {
...
+        super(start, end, connectedTo);
+        this.image = new Image();
+        this.image.src = image;
+        this.image.onload = () => animateCanvas(noUpdateTime);
+        this.size = size;
+        this.opacity = opacity;
+    }
+
+    draw() {
+        const pos = this.getPosition();
```
EliphasNUIT, line 665: missing if (!this.canDraw()) { return;} before getPosition

EliphasNUIT:
Looks great!
You think you could complete all the portals by adding properly sized and colored circle decorations (with UsingSkillMode) at the same position as the icon decorations?

Zerthox:
So two separate decorations? If it's about getting an idea of how large the portals actually are, maybe the icon could resize when  the "use hitbox size" option is enabled?

EliphasNUIT:
Yeah, I was thinking as a separate decoration.
The icon resize is not without interest but I would say that it is an orthogonal evolution to adding circles to represent portals.

Zerthox:
Should portals always render or only on player select? I was thinking the active ones probably should be always displayed since they affect all players. Not sure about inactive portals.

GW2EIEvtcParser/EIData/Colors.cs
```diff
@@ -7,22 +7,29 @@ namespace GW2EIEvtcParser.EIData
 {
     internal class Color
     {
-        private readonly int _r;
-        private readonly int _g;
-        private readonly int _b;
-        private readonly float _a;
+        public int R;
```
EliphasNUIT, line 10: Why public?

EliphasNUIT: public int R {get; set;} would be better in that case, yes

Zerthox:
Some closing notes for future reference: Right now active portals are always displayed where as inactive ones are only displayed on player select. Portal decoration radius is currently 90 units, roughly matching the ingame visual effect size. Actual functional interact size seems to be ~100 units.

PR #766, opened 2023-06-02
Added Magnetic Inversion, Spectral Recall and Lesser Banish Enchantment

Linkaaaaa:
- For Magnetic Inversion it checks the loss of the absorption buff and the visual effect, for logs pre-effcts it's restricted to checking for a damage hit.
- For Spectral Recall, i've restricted it to post december 2018 builds because i don't know the behaviour of the skill was before that. There is a hidden buff 10686 that is lost upon teleporting back or when spectral walk duration runs off. The 10686 buff gets removed 120ms before the Spectral Walk buff is removed, meaning you can safely let the skill run out and it won't trigger Recall.
In this example i've let it run out 3 times, used recall 3 times, ran out once and recalled once

- Added some missing non-crittables
- Added Lesser Banish Enchantments to revenant
- Added Grim Specter buff


EliphasNUIT:
So, is this PR done? Everytime I take a look at it there is a new commit ^^

Linkaaaaa:
Yes i usually add things until i know that you've checked everything :D

EliphasNUIT:
Thanks!

PR #767, opened 2023-06-09
Ai improvements

Zerthox:
Fixes incorrect phase start detections when transition out cast happened after Determined was lost.
Adds a fallback phase to Determined loss when cast not found. Any idea for a better fallback phase name?


EliphasNUIT:
Thanks, the naming's fine!

PR #778, opened 2023-07-13
Removed Power Surge achievement detection

Linkaaaaa:
The achievement notes that it won't be awarded to dead players so shouldn't be an instance buff, was my fault for not reading about it.


EliphasNUIT:
Thanks!

PR #780, opened 2023-07-15
Skill icons and some new buffs

Linkaaaaa:
- Added icons override for old spirit active skills, they got removed from the api, example:

- Added icons for some pet skills
- Added some new buffs
- Renamed some skills and buffs


EliphasNUIT:
Why the XXXBuff naming instead of XXXEffect for certain things and not the others?

Linkaaaaa:
we had decided some time back with zerthox to rename effect to buff to go along with all the buff related classes

EliphasNUIT:
Yes and I am perfectly fine with that. But that does not answer the question, my understanding was that the naming convention would change, not that we would have the two conventions co-existing. Is there a plan to rename all of them?

Linkaaaaa:
I can work on renaming all of them for this pr then

EliphasNUIT:
That'd be perfect!

Linkaaaaa:
done that, should i rename XXXXHit to XXXXDamage too? both are mixed too

EliphasNUIT:
I believe that is more a case by case basis, if something has been wrongly named Hit instead of Damage, yes. But there are a lot of skill ids that represent secondary hits or multi hit components of skills. I don't think you need to go through that in this PR (not sure if it is even necessary).

Linkaaaaa:
Alright then i believe it's good to go, won't touch it unless you notice something wrong

EliphasNUIT:
Thanks!

PR #781, opened 2023-07-16
Fractal Scale

Linkaaaaa:
Sadly don't have logs from when nightmare was scale 100, everything else working correctly as meant to:


EliphasNUIT:
Thanks!

PR #782, opened 2023-07-18
Two instant casts and buffs

Linkaaaaa:
Added instant cast finder for sadistic searing trait and thousand cuts
Added the new heat the soul buff for berserker
Clarified matthias' blood fueled buff naming

This is my last generic pull request for this patch, will work only on the kanaxai's one for the release

### DIFF
--- a/GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
+++ b/GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
@@ -24,7 +24,7 @@ namespace GW2EIEvtcParser.EIData
             new BuffGainCastFinder(SandstormShroudSkill, DesertShroudBuff).UsingDurationChecker(3500),
             // new EXTBarrierCastFinder(DesertShroud, DesertShroud),
             new EXTBarrierCastFinder(SandCascadeSkill, SandCascadeBarrier),
-            new BuffGainCastFinder(SadisticSearing, SadisticSearing),
+            new BuffGainCastFinder(SadisticSearing, SadisticSearing).UsingOrigin(EIData.InstantCastFinder.InstantCastOrigin.Trait),
             new BuffLossCastFinder(SadisticSearingActivation, SadisticSearing).UsingChecker((blcf, combatData, agentData, skillData) => 
             {
                 long sadisticSearingDuration = 10000 - blcf.RemovedDuration;
@@ -33,7 +33,7 @@ namespace GW2EIEvtcParser.EIData
                     return true;
                 }
                 return false;
-            }),
+            }).UsingOrigin(EIData.InstantCastFinder.InstantCastOrigin.Trait),
         };
 
         internal static readonly List<DamageModifier> DamageMods = new List<DamageModifier>


GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
```diff
@@ -21,6 +21,7 @@ internal static class ScourgeHelper
             new BuffGainCastFinder(SandstormShroudSkill, DesertShroudBuff).UsingDurationChecker(3500),
             // new EXTBarrierCastFinder(DesertShroud, DesertShroud),
             new EXTBarrierCastFinder(SandCascadeSkill, SandCascadeBarrier),
+            new BuffGainCastFinder(SadisticSearing, SadisticSearing),
```
EliphasNUIT, line 24: You mean BuffLoss?

GW2EIEvtcParser/EIData/ProfHelpers/Mesmer/VirtuosoHelper.cs
```diff
@@ -29,7 +29,9 @@ internal static class VirtuosoHelper
                 }
                 return true;
             }).WithBuilds(GW2Builds.October2022Balance),
-            new EffectCastFinder(BladeturnRequiem, EffectGUIDs.VirtuosoBladeturnRequiem).UsingSrcSpecChecker(Spec.Virtuoso).WithBuilds(GW2Builds.June2023Balance)
+            new EffectCastFinder(BladeturnRequiem, EffectGUIDs.VirtuosoBladeturnRequiem).UsingSrcSpecChecker(Spec.Virtuoso).WithBuilds(GW2Builds.June2023Balance),
+            new DamageCastFinder(ThousandCuts, ThousandCuts).UsingICD(5000).UsingNotAccurate(true).UsingDisableWithEffectData(),
```
EliphasNUIT, line 33: This should be removed, it is not possible to accurately time the cast time of the skill by damage only, could be wrong up to 5 seconds.

GW2EIEvtcParser/EIData/ProfHelpers/Mesmer/VirtuosoHelper.cs
```diff
@@ -29,7 +29,9 @@ internal static class VirtuosoHelper
...
+            new EffectCastFinder(ThousandCuts, EffectGUIDs.VirtuosoThousandCuts),
```
EliphasNUIT, line 34: Missing UsingSrcSpecChecker(Spec.Virtuoso).

GW2EIEvtcParser/EIData/ProfHelpers/Warrior/BerserkerHelper.cs
```diff
@@ -38,8 +37,8 @@ internal static class BerserkerHelper
             new Buff("Rock Guard", RockGuard , Source.Berserker, BuffClassification.Other, BuffImages.ShatteringBlow),
             new Buff("Feel No Pain (Savage Instinct)", FeelNoPainSavageInstinct, Source.Berserker, BuffClassification.Other, BuffImages.SavageInstinct).WithBuilds(GW2Builds.April2019Balance, GW2Builds.EndOfLife),
             new Buff("Always Angry", AlwaysAngry, Source.Berserker, BuffClassification.Other, BuffImages.AlwaysAngry).WithBuilds(GW2Builds.StartOfLife, GW2Builds.April2019Balance),
+            new Buff("Heat the Soul", HeatTheSoulBuff, Source.Berserker, BuffClassification.OtherConsumable, BuffImages.HeatTheSoul),
```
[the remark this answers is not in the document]
Linkaaaaa: gw612
this buff :D https://wiki.guildwars2.com/wiki/Heat_the_Soul

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
EliphasNUIT: I'm a little bit conflicted about Sadistic Searing, I feel like people seeing it in the rotation would rather think of the burning proc than the fact that the buff was just gained as simply gaining the buff does not mean much, only that a punishment skill was cast, but we already know that and the buff by itself does nothing either.
Manifest Shade is not an instant skill, perhaps there is a way to reliably check the burning proc by checking that the skill was cast?

EliphasNUIT: Only one new custom skillID (for activation, keep application on the normal ID) would have been enough no? Unless you did something else with the buff ID.

Linkaaaaa: haven't done anything else with the normal buff id, i'll do that and commit

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
```diff
@@ -21,6 +24,17 @@ internal static class ScourgeHelper
...
+            new BuffLossCastFinder(SadisticSearingActivation, SadisticSearing).UsingChecker((blcf, combatData, agentData, skillData) => 
+            {
+                AbstractBuffEvent abe = combatData.GetBuffData(SadisticSearing).Where(buffEvent => buffEvent is BuffApplyEvent).LastOrDefault(x => x.Time < blcf.Time);
+                AbstractHealthDamageEvent ahde = combatData.GetDamageData(ManifestSandShadeShadeHit).LastOrDefault(x => x.Time < blcf.Time);
```
EliphasNUIT, line 31: Does the buff disappear only on hit? Or is that also the case when you just summon a shade?

Linkaaaaa: only on hit

GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/ScourgeHelper.cs
```diff
@@ -21,6 +24,17 @@ internal static class ScourgeHelper
...
+                if (abe.Time < ahde.Time && ahde.Time < blcf.Time)
```
EliphasNUIT, line 32: You should null check both abe and ahde.

What is the logic behind the presence of abe? Wouldn't doing ahde = ...LastOrDefault(x => abs(x.Time - blcf.Time) < ServerDelayConstant) and then return ahde != null be enough? Abe could be null if the player starts the log with the buff and buff initials are missing for some reason.

EliphasNUIT: scratch the abs(...) should be blcf.Time - x.Time < ServerDelayConstant to make sure events are properly ordered.

EliphasNUIT:
Thanks!

PR #786, opened 2023-07-21
Skill improvements

Zerthox:
- Portal Interact Cooldown buff (prevents interaction with portals, applied for 2s when taking any portal)
- Level Up skill icon


EliphasNUIT:
Thanks!

PR #789, opened 2023-07-25
Comments and cleanup

Linkaaaaa:
- Compacted GetAgentProfString
- Added comments to parser methods
- Moved file format helpers from ParserHelper to their own class


EliphasNUIT:
Thanks!

PR #792, opened 2023-07-31
More Kanaxai improvements

Zerthox:
- Several new skills/buffs
- Display (presumably) tether-related Aspect buffs
- Categorize Phantasmagoria as debuff
- Display Phantasmagoria & tether link without end
- World Cleaver decoration
- Highlight Aspects when starting tether animated cast
- Fractal Toxic Sickness mechanic


Zerthox:
There's currently a "Kanaxai Something" buff in the buff list, which I would suggest to remove until its purpose is known and we can give it a better name. Also the "Spread AoE" buff applied by Frightening Speed (Numbers) probably doesn't need displaying considering players also have the "Target Order" buffs from 1-5 simultaneously.

Linkaaaaa:
if you don't simulate the buff it can't be used for the decoration, target order is removed way later than the buff i've added

EliphasNUIT:
> There's currently a "Kanaxai Something" buff in the buff list, which I would suggest to remove until its purpose is known and we can give it a better name. Also the "Spread AoE" buff applied by Frightening Speed (Numbers) probably doesn't need displaying considering players also have the "Target Order" buffs from 1-5 simultaneously.

That buff can be renamed, I forgot to get rid of it while looking at what it was supposed to do.

> if you don't simulate the buff it can't be used for the decoration, target order is removed way later than the buff i've added

That's is only for buff related APIs on AbstractActors, CombatData.GetBuffData APIs can still be used in decoration computations.

Zerthox:
I'll just throw this in here as well: maybe we should move all of the encounter-specific buffs to their respective encounter logic at some point.

Zerthox:
> That is only for buff related APIs on AbstractActors, CombatData.GetBuffData APIs can still be used in decoration computations.

So should the decoration logic be rewritten to not rely on player buff status or do we want to keep the buff with an appropriate name & icon?

EliphasNUIT:
That's two different problems, you should use the best possible APIs for a given situation. If you are worried about "useless" buff appearing on graphs and tables, I can think of a way to allow simulating a buff without streaming it to JSON/HTML.

EliphasNUIT:
> 

This is something I thought about, but quite a few fights share the same buffs which means we would have to verify that buffs we would be moving are not also present elsewhere (I assume by moving the buffs to EncounterLogic, you would also be adding only the buffs from the relevant logic to the container? Otherwise, you can forget my remark.)

EliphasNUIT:
I did add a "Hidden" BuffCategory last night, feel free to check it up and maybe move certain buffs in there if you believe them to be redundant.

Zerthox:
> > I'll just throw this in here as well: maybe we should move all of the encounter-specific buffs to their respective encounter logic at some point.
> 
> This is something I thought about, but quite a few fights share the same buffs which means we would have to verify that buffs we would be moving are not also present elsewhere (I assume by moving the buffs to EncounterLogic, you would also be adding only the buffs from the relevant logic to the container? Otherwise, you can forget my remark.)

I would keep buffs shared between encounters there or maybe in cases like Instabilities move them to the parent (Fractal) encounter logic. I think the most important difference between always adding all of them and only adding relevant ones is unique names. Right now unique names are enforced globally across encounters, which might or might not be desirable to keep?

EliphasNUIT:
I see the point regarding naming but I don't think we have much choice to begin with, logging whole instances is possible so we could only "fix" the naming for bosses that obey the "one instance, one boss" rule but even that could be problematic if the encounter has a CM mode with different buffIDs using the same name and the loggers switch between. 
We also have to make sure that we are not breaking non officially supported logs, say if you add FractalInstabilities to encounters inheriting from FractalLogic, that would remove instabilities from non supported fractal logs.

Zerthox:
Makes sense. So if we decide to move some of them, we should still add all of them to the container.

EliphasNUIT:
Thanks!

PR #794, opened 2023-08-01
Skill decorations

Zerthox:
Building upon #761 this adds skill decorations for several ground targeted player skills to the combat replay. Skills in **bold** are drawn always.

- Engineer: **Defense Field**
- Guardian: Ring of Warding, Line of Warding, Sanctuary, **Wall of Reflection**, Shield of the Avenger, **Valiant Bulwark (Firebrand)**
- Mesmer: **Feedback**, **Veil**, Null Field, Rain of Storms (Virtuoso), Thousand Cuts (Virtuoso)
- Necromancer: Corrosive Poison Cloud, Plaguelands, **Well of Blood**, Well of Suffering, Well of Darkness, Well of Corruption, Shade (Scourge)
- Ranger: **Hunker Down (Siege Turtle)**, **Sublime Conversion (Druid)**
- Revenant: **Inspiring Reinforcement**
- Thief: **Seal Area**, **Shadow Refuge**
- Warrior: **Winds of Disenchantment (Spellbreaker)**


GW2EIEvtcParser/ParsedData/CombatEvents/StatusEvents/EffectEvent.cs
```diff
@@ -18,6 +19,9 @@ public abstract class EffectEvent : AbstractStatusEvent
 
         public long Duration { get; protected set; }
 
+        /// <summary>The effect's rotation around the z-axis in degrees</summary>
+        public float Rotation => RadianToDegreeF(Orientation.Z);
```
EliphasNUIT, line 23: RotationZ would be a better naming.

EliphasNUIT: Sounds good to me!

Zerthox:
Necromancer Wells and Scourge Shades are commented out currently. Null Field probably falls into that group as well. Should we include these skills or leave them out for now and maybe add them in the future?

EliphasNUIT:
I don't see any reason why those couldn't be shown when the player is selected. For shades you could decide if they should be visible by default or not based on if the scourge provides alacrity or not.
Anything that is not relevant for every squad member to see shouldn't be visible by default. 
Skills which main purpose is to provide support, like projectile destruction, providing "rare" buffs like stab, aegis, alac, quickness and stealth or removing conditions. Skills that are cast purely for dps or skills that have some small utilities despite the damage should be visible on selection.
For example: well of precognition -> always visible, well of action -> visible on selection.

EliphasNUIT:
Thanks!

PR #797, opened 2023-08-03
Initial overhead decorations and related additions

Linkaaaaa:
- Target orders decorations
- Power of the void decorations
- Squad target marker guids
- Parser icons for overhead decorations
- New GetBuffStatus overload for multiple buffs array


EliphasNUIT:
Looks good! Thanks.

PR #809, opened 2023-09-05
Cerus mechanics

Linkaaaaa:
Added various cerus mechanics and moved exposed and debilitated to general strike missions logic


EliphasNUIT:
Thanks!

PR #810, opened 2023-09-15
Combat Replay Arenas for Secret of the Obscure strike missions

LunusLupus:
- Added CR arena images for SotO strikes
- Adjusted Cosmic Observatory CR arena position
- Adjusted Temple of Febe CR arena position


EliphasNUIT:
Thanks!

PR #814, opened 2023-10-06
Some new buffs + dhuum 10% mechanic

Linkaaaaa:
Added 2 wvw buffs, pure oblivion to dhuum and grim specter to necro


EliphasNUIT:
Thanks!

PR #824, opened 2023-10-29
Fix Arkk Horizon Strike decorations

Zerthox:
Add cap to non-effect-based Arkk Horizon Strike decorations using Determined & Stun apply.


EliphasNUIT:
Thanks!

PR #829, opened 2023-11-17
Fix request URL for DPS.Report upload

Nightara:
Spaces in the get string cause DPS.Report to return a malformed (Non-JSON) response, which crashes the application.


EliphasNUIT:
Thanks

PR #836, opened 2023-12-08
Fix duplicate Arkk Bloom phases

Zerthox:
Some of the Blooms can spawn delayed sometimes. I had one case myself with a 21ms difference. I upped the max difference from 10ms to 100ms to hopefully catch all of those.


EliphasNUIT:
Thanks!

PR #838, opened 2023-12-11
Fix Skorvald split phase start/end times

Zerthox:
Skorvald split phases were starting with first Anomaly combat enter rather than when the boss got invulnerable.


Zerthox:
I just noticed it's the same behavior on splits phases in other encounters as well. That behavior does not make sense to me. You wouldn't start a boss phase when the first player hits the boss, right? Ideally a split phase should start when the adds spawn and end when they are dead. Or alternatively start when the boss can't be attacked anymore and end when it's attackable again.

But considering this is a thing across a lot of encounters I will close this for now and we'll discuss it.

EliphasNUIT:
But that is already the case? In the case of anomalies we take the max between FirstAware/Spawn/EnterCombat based on the presence of the events, we don't check incoming/outgoing damage events. For end time it is the min between LastAware/Dead event.
All of those are clamped inside the initial (start, end) that was computed.

PR #840, opened 2023-12-12
Added dhuum's messenger fixation mechanic and decoration

Linkaaaaa:


GW2EIEvtcParser/EncounterLogic/Raids/W5/Dhuum.cs
```diff
@@ -65,7 +65,17 @@ public Dhuum(int triggerID) : base(triggerID)
...
+            new PlayerDstBuffRemoveMechanic(EchosPickup, "Freed from Echo", new MechanicPlotlySetting(Symbols.Square,Colors.Blue), "F Echo","Freed from Ender's Echo", "Freed from Echo", 0).UsingChecker( (br,log) => !log.CombatData.GetDeadEvents(br.To).Where(x => Math.Abs(x.Time - br.Time) <= 150).Any()),
+            new PlayerSrcBuffApplyMechanic(DhuumsMessengerFixationBuff, "Messenger Fixation", new MechanicPlotlySetting(Symbols.CircleOpenDot, Colors.Brown), "Mess Fix", "Fixated by Messenger", "Messenger Fixation", 10).UsingChecker((bae, log) =>
+            {
+                // Additional buff applications can happen, filting them out
+                AbstractBuffEvent firstAggroEvent = log.CombatData.GetBuffDataByDst(bae.To).Where(x => x.BuffID == DhuumsMessengerFixationBuff).FirstOrDefault();
+                if (firstAggroEvent != null && bae.Time > firstAggroEvent.Time && bae.Initial)
```
EliphasNUIT, line 73: firstAggroEvent.Time + ServerDelayConstant.
In the rare occasion where we would only get the first buff initial and not the traditional apply event, this check would filter out it as bae would be equal to firstAggroEvent and the time check will always fail

EliphasNUIT:
Thanks!

PR #842, opened 2023-12-17
Offset Artsariiv Beaming Smile decorations

Zerthox:
Offset Beaming Smile decorations at Artsariiv to be closer to ingame attack telegraph. 60 units is an educated guess based on POVs. Someone would have to measure it for a precise offset.


EliphasNUIT:
Thanks!

PR #850, opened 2023-12-30
Updated achievement checks for various encounters

Linkaaaaa:
- Created SetOnPlayerCustomInstanceBuff and streamlined the usage within SetInstanceBuffs to reduce code repetition
- Created CustomCheckTakingTurns and removed its instance buff, instead moved it to PlayerStatusMechanic<DeadEvent>
- Created CustomCheckManipulateTheManipulator for logs that don't have any present achievement buff
- Small cleanups for Qadim

### DIFF
--- a/GW2EIEvtcParser/EncounterLogic/FightLogic.cs
+++ b/GW2EIEvtcParser/EncounterLogic/FightLogic.cs
@@ -6,6 +6,7 @@ using GW2EIEvtcParser.Exceptions;
 using GW2EIEvtcParser.Extensions;
 using GW2EIEvtcParser.ParsedData;
 using static GW2EIEvtcParser.ArcDPSEnums;
+using static GW2EIEvtcParser.ParserHelper;
 using static GW2EIEvtcParser.EncounterLogic.EncounterLogicUtils;
 using static GW2EIEvtcParser.EncounterLogic.EncounterLogicPhaseUtils;
 using static GW2EIEvtcParser.EncounterLogic.EncounterLogicTimeUtils;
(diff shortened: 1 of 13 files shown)


GW2EIEvtcParser/EncounterLogic/FightLogic.cs
```diff
@@ -422,5 +423,29 @@ internal virtual void EIEvtcParse(ulong gw2Build, FightData fightData, AgentData
...
+        /// The buff must be present on any player at the end of the encounter.<br></br>
+        /// </summary>
+        /// <param name="log">The log.</param>
+        /// <param name="buff">The buff ID to add.</param>
+        /// <param name="stack">Amount of buff stacks (0-99).</param>
+        /// <returns>
+        /// A <see cref="List{T}"/> containing a <paramref name="buff"/> and its <paramref name="stack"/> if present, otherwise empty.<br></br>
+        /// To be used to add as range to <see cref="InstanceBuffs"/>.
+        /// </returns>
+        internal static List<(Buff, int)> SetOnPlayerCustomInstanceBuff(ParsedEvtcLog log, long buff, int stack = 1)
```
EliphasNUIT, line 437: Protected static would be better, this does not need to be called outside of FightLogic and its children.
Also the method would be better named "GetXXX" instead of "SetXXX" now that it returns a list instead of interactive with one.

Linkaaaaa: Should i change List to IReadOnlyList too?

EliphasNUIT: Not a Public API so not mandatory but how it is used, it would make sense

EliphasNUIT:
Thanks!

PR #864, opened 2024-02-09
Added TryGetFirstAgentItem API

Linkaaaaa:
- Added TryGetFirstAgentItem for all enum types and general int
- Updated the code where the new api was usable


EliphasNUIT:
Thanks.

PR #875, opened 2024-04-01
Cerus adjustments

Linkaaaaa:
- Improved the Permanent Embodiments lifespan, including 3500ms spawn and despawn time windows, hard coded the cast durations in order to not rely on correct readings by arcdps.
- Fixed Envious Gaze beam appearing after Cerus gains a breakbar. The issue was Cerus gaining the breakbar in the time frame inbetween the beam indicator and the beam starting to rotate.
- Fixed the Permanent Embodiments mechanics not getting cancelled at the start of a split phase.
- Added various comments.


GW2EIEvtcParser/EncounterLogic/Strikes/SotO/TempleOfFebe.cs
```diff
@@ -500,33 +500,119 @@ internal override void ComputeEnvironmentCombatReplayDecorations(ParsedEvtcLog l
...
             bool startTrimmed = false;
-            foreach (AbstractCastEvent e in castEvents)
+
+            // Cast durations of each mechanic with +- 20ms leeway
+            foreach (AbstractCastEvent cast in castEvents)
             {
+                int castDuration = 0;
+                switch (cast.SkillId)
+                {
+                    case CrushingRegretNM:
```
EliphasNUIT, line 523: Target species ID check would look easier on the eyes no? Would also make it possible to do it outside of the forEach.
Or is it possible for Embodiments to cast different skills? Like Despair casting Crushing Regret?

Linkaaaaa: It's not possible but also i've included normal mode skills which makes no sense, this is only for cm :D
Since we already have a target switch before calling the method, i could pass the cast duration in the parameters and discard the switch completely

Linkaaaaa: Should be better now

EliphasNUIT:
Thanks!

PR #876, opened 2024-04-03
Small fixes

Linkaaaaa:
- Fixed the combat replay timeline resetting back to 0 instead of going backwards when it reaches the end.
- Improved tooltip for the players when hovering the names
- Added Weight of the World to the damage modifiers
- Updated the description of the modifiers being additive and multiplicative


EliphasNUIT:
Thanks!

Issue #772, opened 2023-06-22
Unknown Skills Icons


Drevarr:
The following skills are listed as unknown in `Player Healing Summary>Outgoing Healing`.  As discussed on Discord, some may be due to differing versions of the Healing-Stat addon. However as example, `Med Blaster` was unknown on half the wvw fights on 6/21 and I was the only heal scrapper in squad with Healing-Stats Addon,

Request a review for potential override methodology to ensure proper identification.

| ID | Current Name | Proposed Skill Name | Current Icon | Proposed Icon |
|---:|--------------|---------------------|--------------|---------------|
| 29997 |  UNKNOWN  | Healing Prism | [Healing_Prism](https://wiki.guildwars2.com/images/f/f4/Healing_Prism.png) |  |
| 30235 |  UNKNOWN  | Med Blaster | [Unknown](https://render.guildwars2.com/file/1D55D34FB4EE20B1962E315245E40CA5E1042D0E/62248.png) | [Med_Blaster](https://wiki.guildwars2.com/images/6/65/Med_Blaster.png) |
| 43759 |  UNKNOWN  | Sand Cascade | [Sand_Cascade](https://wiki.guildwars2.com/images/1/1e/Sand_Cascade.png) |  |
| 40774 |  UNKNOWN  | Sand Flare | [Sand_Flare](https://wiki.guildwars2.com/images/f/f0/Sand_Flare.png) |  |
| 30564 |  UNKNOWN  | Soothing Detonation | [Unknown](https://render.guildwars2.com/file/1D55D34FB4EE20B1962E315245E40CA5E1042D0E/62248.png) | [Soothing_Detonation](https://wiki.guildwars2.com/images/a/ad/Soothing_Detonation.png) |
| 5981 |  UNKNOWN  | Turret Heal | [Unknown](https://render.guildwars2.com/file/1D55D34FB4EE20B1962E315245E40CA5E1042D0E/62248.png) | [Healing_Turret](https://wiki.guildwars2.com/images/4/42/Healing_Turret.png) |
| 13980 |  UNKNOWN  | Windborne Notes | [Windborne_Notes](https://wiki.guildwars2.com/images/8/84/Windborne_Notes.png) |  |

Two logs that contain the above skills:
[WvW_Logs.zip](https://github.com/baaron4/GW2-Elite-Insights-Parser/files/11828690/WvW_Logs.zip)

[...]

PR #879, opened 2024-04-11
Added varius skill decorations, icons and buffs

Linkaaaaa:
# Decorations
- Signet of Mercy
- Illusion of Life
- Unstable Bladestorm
- Signet of Undeath
- Spectral Ring
- Citadel Bombardment
- Coalescence of Ruin
- Drop the Hammer
- Eternity's Requiem

# Misc
- New Kalla skill icons for the reworked summons
- Nature's Renewal icon for #772 
- Renamed some SkillIDs to distinguish between the skill and the buff
- Added Photon Barrier for Holosmith
- Added Sky-Chak Striker ranger pet
- Added FindCentralPoint API to find the central 3d position given a list of positions


GW2EIEvtcParser/EIData/ProfHelpers/Revenant/RevenantHelper.cs
```diff
@@ -274,14 +274,67 @@ internal static void ComputeProfessionCombatReplayActors(AbstractPlayer player,
...
+                        var positions = new List<Point3D>();
+                        foreach (EffectEvent hitEffect in eternitysRequiemHits.Where(x => x.Time >= effect.Time && x.Time <= effect.Time + 2800 && x.Dst == null))
+                        {
+                            positions.Add(hitEffect.Position);
+                            (long, long) lifespanHit = hitEffect.ComputeLifespan(log, 1000);
+                            var connector = new PositionConnector(hitEffect.Position);
+                            replay.Decorations.Add(new CircleDecoration(120, lifespanHit, color, 0.5, connector).UsingFilled(false).UsingSkillMode(skill));
+                        }
+                        (long, long) lifespan = (effect.Time, effect.Time + 2800);
+                        var centralPosition = Point3D.FindCentralPoint(positions);
```
EliphasNUIT, line 303: I'm quite confused about this, the projectiles look quite random to me, is there a documentation somewhere that indicates that the average of the 9 impacts point is the center of the AoE?
Same remark for citadel bombardment, the impacts don't look to be equally distributed in every direction.

GW2EIEvtcParser/EIData/ProfHelpers/Revenant/RenegadeHelper.cs
```diff
@@ -63,5 +65,57 @@ internal static bool IsKnownMinionID(int id)
...
+                                // Hit circles
+                                (long, long) lifespanHit = hitEffect.ComputeLifespan(log, 500);
+                                var connector = new PositionConnector(hitEffect.Position);
+                                replay.Decorations.Add(new CircleDecoration(120, lifespanHit, color, 0.5, connector).UsingFilled(false).UsingSkillMode(skill));
+                            }
+                            // AoE Radius and icons
+                            (long, long) lifespan = (effect.Time, effect.Time + 3000);
+                            var centralPoint = Point3D.FindCentralPoint(positions);
+                            var centralConnector = new PositionConnector(centralPoint);
+                            replay.Decorations.Add(new IconDecoration(ParserIcons.EffectCitadelBombardmentPortal, CombatReplaySkillDefaultSizeInPixel, CombatReplaySkillDefaultSizeInWorld, 0.5f, lifespan, playerConnector)
```
EliphasNUIT, line 110: The icon will overlap with the icon of the player. Moreover, the portal does not follow the player, you can move around during the cast, the portal will remain at the position where the cast is finished

GW2EIEvtcParser/EIData/ProfHelpers/Revenant/RevenantHelper.cs
Linkaaaaa: Both skills have a radius where the impacts will land, but neither have a displayed aoe.
I was using the player position when the first effects appears, but for eternity's requiem it would cause it to be offset while in movement https://gyazo.com/f9c93cd7bb24e9b43d782e92912fd8d3
Here is how it looks with an extimated central point https://gyazo.com/307784b810dfd63f26d25fdfd82b9248
For both skills the actual damage is coming from the small rings, the large one is just as a general indicator to place the skill icon inside

EliphasNUIT: I understand what you are trying to achieve, the part I'm conflicted about is the accuracy of the methodology. How accurate is getting the middle of all impacts compared to the actual center of the AoE?
It feels like it would be better to show only an icon to indicate a general position of where the AoE is and omit the circle for the AoE. Looking at it in game right now, it really does not feel like the average of the impacts is the center of the AoE.

Linkaaaaa: > I was using the player position when the first effects appears
And for citadel bombardment it would cause the ring to be at the portal location, so i swapped that to be the portal image and created the new position to locate the skill icon

Linkaaaaa: This is how it looks like without the ring if you prefer it https://gyazo.com/c139c81365eea1e6fc9aeb7ea5001778

EliphasNUIT: Yes, comment the circle for now. We can add it later if we find a good way to show that the circle is approximative, as of today people would take it as truth which would be incorrect.

Linkaaaaa: We could have dashed/dotted lines for approximations, like the line decoration we were talking about the other day, but I don't know how that information would be understood by the user though

EliphasNUIT: That depends, dashed lines were proposed to avoid bad overlaps for line decorations and not for approximations, dashed lines being approximation for certain stuff and not others would not work out nicely.

PR #880, opened 2024-04-18
Fix incorrect phase names due to filtering

Zerthox:
This fixes incorrect phase names for phases after a < 2s phase, which can theoretically occur for several encounters. The problem was that `GetPhasesByInvul` filtered < 2s phases already, resulting in encounter logic implementations assigning incorrect phase names since some of them assume all phases to be present.

### DIFF
--- a/GW2EIEvtcParser/EncounterLogic/EncounterLogicPhaseUtils.cs
+++ b/GW2EIEvtcParser/EncounterLogic/EncounterLogicPhaseUtils.cs
@@ -74,7 +74,7 @@ namespace GW2EIEvtcParser.EncounterLogic
             {
                 phases.Add(new PhaseData(last, end));
             }
-            return phases; // skip filtering too short phases here, it may mess with phase names
+            return phases.Where(x => x.DurationInMS > 100).ToList(); // only filter unrealistically short phases, otherwise it may mess with phase names
         }
 
         internal static List<PhaseData> GetPhasesByInvul(ParsedEvtcLog log, long skillID, AbstractSingleActor mainTarget, bool addSkipPhases, bool beginWithStart, long start, long end)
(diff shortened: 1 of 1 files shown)


EliphasNUIT:
Do you know which fights we are talking about and the order of magnitude of the filtered phases? 
There has been occurrences in the past where buffs we use to determine phase could be dropped and regained on super small windows and that is why that duration check was introduced there. We could reduce it to something more reasonable as I don't think phases that are around 50/100ms long can be considered as legitimate phases while a phase with a duration of 1s could be.

Zerthox:
In theory it can happen on every fight where we do phase names based on the index of phases returned by `GetPhasesByInvul`. I listed all of them below. It becomes extra problematic if the logic uses `i % 2` to distinguish between boss and split phases. Boss phases can end up being named as if they were a split phase and the other way around. I marked encounters where this could theoretically happen with \*.
In practice it's only the encounters where it is possible to go < 2s in any phase. This should definitely be possible on MAMA, Siax, Artsariiv and I reckon it might be possible on some Raid/Strike bosses where you can have really fast splits (e.g. oneshotting adds with Bladesworn stack). For example Adina speedkills would trigger the bug if the Adina encounter logic didn't implement the Invulnerability phase logic manually.

- Raids
  - Vale Guardian \*
  - Gorseval \*
  - Sabetha \*
  - Escort \*
  - Samarog \*
  - Conjured Amalgamate \*
  - Qadim \*
- Fractals
  - Generic Fractal encounter logic
  - MAMA \*
  - Siax \*
  - Ensolyss  \*
  - Skorvald \*
  - Artsariiv \*
  - Arkk
  - Elemental Ai
  - Kanaxai
- Strikes
  - Freezie \*
  - Shiverpeaks Pass
  - Voice and Claw
  - Aetherblade Hideout \*
  - Xunlai Jade Junkyard
  - Kaineng Overlook
  - Old Lions Court
  - Temple of Febe \*
- Other
  - Soo Won (Open World)
  - Mordremoth (Story)

Zerthox:
I think filtering with a lower millisecond duration is the sensible thing to do.

PR #894, opened 2024-06-04
Initial Eparch CM encounter support

Zerthox:
Adds primitive encounter support for Eparch CM:

- [x] Main, secondary & trash targets
- [x] Phases
- [x] Encounter buffs


GW2EIEvtcParser/ParserHelpers/SkillIDs.cs
```diff
@@ -4270,6 +4270,21 @@ public static class SkillIDs
         public const long RelicOfTheTwinGenerals = 72707;
         public const long Panopticon = 72843;
         public const long StaringVoid = 72851;
+        public const long RageEmpowerment = 72399;
```
EliphasNUIT, line 4273: IDs should be properly ordered.

Zerthox:
Still missing a (most likely HP-based) check to distinguish CM from normal mode logs.

EliphasNUIT:
Looks good thanks, I'll see for the encounter icons + encounter specific damage mods.

PR #896, opened 2024-06-12
Format numbers for better readability

Zerthox:
This attempts to change number formatting for better readability.
Current format is space as thousand separator and `.` as decimal separator like `1 234 567.89`.

Alternatively, we could also shorten to `1.23M` and `45.6k` and display the full number as tooltip on hover.


Zerthox:
Closing in favor of #1168.

PR #902, opened 2024-06-15
Fix soo won crash due to arcdps not logging effect

Linkaaaaa:
Fixes the 2 soo won logs not parsing from discord due to arcdps not logging the orb effects


EliphasNUIT:
I'll also improve the TryGets on my side, currently they return true despite not finding anything.
Thanks.

PR #912, opened 2024-06-29
Fix down contribution support in WvW post lastDownBefore90 event removal

Ariacell:
Small PR to re-enable support for parsing down contribution in detailed WvW logs now that health update events are available in the limited WvW set as of Arc evtc build 20240529. The method appears correct when comparing with in-game arcdps numbers for down contribution, please let me know if you have other tests you normally like to run against this sort of calculation change.

Note down contribution in non-detailed WvW is unchanged (all zeroes as that is current release behaviour).
I've attached some sample manual test scenarios below for logs from both before and after the change to confirm backwards compatibility.

Sample Offensive Stats table with log from Pre EVTCBuild 20240529
image
[PreEVTCBuildSample.zip](https://github.com/user-attachments/files/16038273/PreEVTCBuildSample.zip)

Post-EVTCBuild 20240529
image
image
[PostEVTC20240529LogSample.zip](https://github.com/user-attachments/files/16038265/PostEVTC20240529LogSample.zip)


EliphasNUIT:
Looks good, thanks!

PR #913, opened 2024-06-30
Fix thief venom stacks capacity

Linkaaaaa:
They got increased in the 25/06 patch
https://wiki.guildwars2.com/wiki/Game_updates/2024-06-25#Thief
[venoms.zip](https://github.com/user-attachments/files/16043896/venoms.zip)


EliphasNUIT:
If there is something wrong with BuffInfoEvent.MaxStack it is better to report it to ArcDPS than try to do a by-pass.

Linkaaaaa:
There's nothing wrong, the new buff capacity is logged correctly

EliphasNUIT:
Then everything is working as expected no?

Linkaaaaa:
Yes they work but do we want to keep the incorrect number there instead of splitting it with a game build?

EliphasNUIT:
That is what we have been doing since the introduction of BuffInfoEvent half a decade ago, the burden of maintenance of capacity was shifted to it.

Linkaaaaa:
Well nevermind then

PR #925, opened 2024-08-26
Add secondary targets

Zerthox:
This adds some "progress-blocking" split phase DPS targets as secondary targets to the main phase (akin to Kaineng Overlook).

Fractals:
- Knights at MAMA (3 targets)
- Echos at Siax (8 targets)
- Anomalies at Skorvald (8 targets)
- Archdiviner & Gladiator at Arkk (2 targets)
- Incarnations at Eparch (5 targets)

Raids:
- Split Guardians at Vale Guardian (6 targets)
- Souls at Gorseval (8 targets)
- Kernan, Knuckles & Karde at Sabetha (3 targets)
- Guldhem at Samarog (2 targets)
- Hands at Adina (10 targets)

### DIFF
--- a/GW2EIEvtcParser/EIData/Actors/AbstractActor.cs
+++ b/GW2EIEvtcParser/EIData/Actors/AbstractActor.cs
@@ -85,6 +85,11 @@ namespace GW2EIEvtcParser.EIData
             return AgentItem.IsSpecies(id);
         }
 
+        public bool IsAnySpecies(IEnumerable<int> ids)
+        {
+            return AgentItem.IsAnySpecies(ids);
+        }
+
         public bool IsAnySpecies(IEnumerable<ArcDPSEnums.TrashID> ids)
         {
             return AgentItem.IsAnySpecies(ids);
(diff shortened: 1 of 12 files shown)


Zerthox:
One open question would be whether we should be merging "duplicate" targets, for example the 2 instances of Red/Green/Blue Guardian at Vale Guardian.

Linkaaaaa:
Number them like we do for Tormented Deads, Anomalies and others

https://github.com/baaron4/GW2-Elite-Insights-Parser/blob/master/GW2EIEvtcParser/EncounterLogic/Raids/W5/SoullessHorror.cs#L127-L132

EliphasNUIT:
From the code the only agents that would need renaming would be the Gorseval ones, others are already renamed with numbers or cardinal positions.

EliphasNUIT:
For Adina, you could do a 
phases[0].AddSecondaryTargets(splitPhase.Targets) in GetPhases

Zerthox:
That would mean they only get added to the main phase when phases are being computed, right? Is it fine to have different behavior for the main phase targets depending on whether other phases are computed?

Zerthox:
Health check is a bit problematic since Adina can easily go below 25% before the 3rd split starts. In the log I am testing with she gains Determined nearly 1s after a < 25% HP update. And Bladesworn Slashes can hit for like 250k, good timing on one could push her below 23% going into the split.

Technically we could check < 23% as the Phase 4 Hands are not supposed to spawn before 20%. Alternatively I could try to do it based on Determined applies instead. They are needed for phase computation below anyway, only need to move it up.

Zerthox:
Should I add the "optional" hands as secondary targets to Phase 4?

EliphasNUIT:
Determined check could fail on late start logs though and add actual P4 hands as secondary targets to the main phase.
If last Determined event is an apply, use log.FightData.FightEnd, if last determined event is a remove, us event.Time.

Zerthox:
We can go back to checking <23% health if thats more robust?

EliphasNUIT:
Thanks!

PR #926, opened 2024-08-30
Fix globules without position

Zerthox:
Fix for globules with position events that are after log end & adds a null check.


EliphasNUIT:
Thanks!

PR #927, opened 2024-09-01
Add instant cast detection for Renegade Band Together

Zerthox:
This adds instant cast finders for the enhanced versions of the Legendary Renegade skills. Initially wanted to to it based on the buff, but turns out it's POV only. Using effects instead now, for the instant casts they occur at the same time as the buff remove.


EliphasNUIT:
Thanks.

PR #930, opened 2024-10-08
Split the executalbe into two projects, `Parser` (windowed, name unchanged) and `ParserCLI`.

SaculRennorb:
Split the executalbe into two projects, windowed (old) and cli.
Moved Settings into ParserCommon to share it between cli and windowed. Changed the test dependencies to ParserCommon which allows the tests to also target non windows frameworks.

Updated the language version of subprojects to 10 for upcoming changes.
Had to change the .editorconfig, because a var declaraion was in violation of an error rule. Probably a change in what is considered for the specific rule between versions.
Moved the HighDPI setting form the manifest into the project file.
Some of the properties in the windowed project file also got shuffled around, i tried to group them in regards of platform features should we ever want to try and give the unified version another shot. 

I really tried to make this work as one codebase, but VS _really_ doesn't want you to build two different executables form the same codebase. I again almost got it working, but in the end the Forms designer wouldnt load. If you ever want to attempt the unification you need to
 - change it to a target framework set instead of a specific one
 - set different output paths
 - set different intermediate paths before the actual project file gets loaded. This requires adding a Build.Directory.props (iir the file name correctly)
 - splitting project packages and includes based on target framework (`<Compile Remove="" />` forms and resources in the generic version)
 - find a way to get the forms designer working (i got it to open but it would just time out while trying to load)

I did some preliminarily testing and everything seems to work just as before, except you need a new version of the runtime ofcourse. 
I also don't know your release (pipeline) and **you might need to adjust some attributes in the CLI project** regarding updating etc.


GW2EIParserCLI/GW2EIParserCLI.csproj
```diff
@@ -0,0 +1,35 @@
...
+        <OutputPath>..\GW2EI.bin\$(Configuration)\</OutputPath>
+        <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <None Include="Content\ArenaNet\LI.ico" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="..\GW2EIDiscord\GW2EIDiscord.csproj" />
```
EliphasNUIT, line 31: Unless I missed it, I don't see any discord related stuff in GW2EIParserCLI?

SaculRennorb: Absolutely correct. It was just copy pasted from the main one.

EliphasNUIT:
Thanks!

PR #933, opened 2024-11-05
[WIP] Perf. 

SaculRennorb:
Ok, this is taking way longer than expected. I'm also somewhat busy with other things right now, so I'll just open this pr for scrutinization.

This pr also:
- Adds some tests (without any logs, see `GW2EIParser.tst\ParsingErrors.cs` and `GW2EIParser.tst\ParsesWithoutError.cs`.
  I have not fully tested everything after the final rebase, so more commits might follow fixing changes from that.
- Introduces Nullability. In general its a very good idea to use it, but i'm not sure if i will find the time to resolve the remaining related issues correctly.
- updates to C#12
- switches to file scoped namespaces
- enabled global default usings

This means it touches basically every file, but don't be afraid: just tell git to ignore whitespaces, that already cuts out 85% of the changes.

### Regarding the numbers:
- This improves overall processing speed by 15% in average up to 30% for simulation heavy logs (cerus). I'm still working on better benchmarks
- I have not measured memory usage yet, but it should be equal or less than before since i mostly just removed copies and streamed iterations.
- For some unknown reason the output files are ~0.5% larger. I have not figured out why that is, but it could very well just be slight changes in compression efficiency because of changes in numeric output (we now omit tailing zeros after the dot)

I should also mention that there are some changes like the the explicit naming of variables that can easily be changed back in case you prefer it that way, but for now i left everything the way i was working with it.

I will also paste some notes i wrote down during the whole process, and while you didn't ask for a code review i necessarily read a lot of the codebase so it also has some recommendations at the end:

### Some notes on genral changes and preformance optimizations:

- Liberal use of the `var` keyword:
  This allows easier itteraton, because of the following scenario:
  ```cs
  int calculate(int a, int b) { /**/ }

  int result = calculate(1, 3);
  do_something_else(result);
  ```
  If calculate needs to be changed to return a float, all usages also need to be changed because even just storing the temporary result now fails (on in this case even worse; truncates the result).
  If the code was written as
  ```cs
  int calculate(int a, int b) { /**/ }

  var result = calculate(1, 3); // <-  var use here
  do_something_else(result);
  ```
  Then `calculate`'s return type can be changed freely without causing type errors on every usage. This ofcorse entails that 
  `do_somethign_else` also works with the new type, for for things like numerics and enumerable types this is usually the case.  
  The only conceivable downside comes up in the case where the new return type does convert into the next input type, but does so unexpectedly. In practice this happens so rarely that it is vastly outweight by the flexibility gained for refactoring.

- Replacing `List<T> myList = new List<T>();` (or similar) with just `List<T> myList = new()` or preferably `var myList = new List<T>()` in local contexts:  
  This just makes hte code more readable. If hte type is already stated in the same line, there is no need to repeat it again.  
  Also again this makes it easier to refactor, because only one change is needed instead of two.

- File level namespaces:  
  Namespaces can now be global for a file without needing to wrap the classes etc in braces and add an indentation level.

- Replacing `new List<T>()`, `new T[]` and similar with just `[ ]` (and to a lesser extend `[ ..other]`):
  This is a new C#12 feature that allows collection initialization regardless of the collection type. This has two main upsides, but unfortunately also one downside:
  - This can generate more specific types. for example, doing things like
    ```cs
    IEnumerable<T> get_elements() => [ ];
    ```
    generates access to `Array<T>.Empty`, which is a very efficient type to use in this scenario, as it only gets allocated once for the whole program. Using `=> new List<T>()` instead allocates every time the function is called.
  - It makes it easier to refactor, since hte target type can change without the semantics having to change.
  - The downside is that
    ```cs
    var list = new List<T>();
    list.AddRange(readOnlyList);
    ```
    is still more efficient for anything that is not a list as the input type. There is an optimization specifically for `List<T> newList = [..oldList];` specifically when oldList is the plain `List<T>` type, but it does not yet work for even `IReadOnlyList`. For al other types this spread initialization will enumerate the collection without reserving space in the target first, so if you know the oldList length it is always better to explicitly reserve first before adding the elements.
  - I also didn't do it in places where there probably should be a capacity set, but I either don't know it or haven't gotten to setting it.

- Replaced `Newtonsoft.Json` with `System.Text.Json`:
  Dotnet now has its own json library, and its really good. It causes almost no allocations and is therefore very fast and memory efficient.  
  It also is a bit smarter with how numeric output is produced, and I would have expected this to also reduce the output size, but it somehow increased the final output by ~0.5% (after compression) for some reason.

- Changed several interfaces from `IReadOnlyList` to `IEnumerable`:
  This is relevant for collections that get filtered and transformed without being used as the list itself. In almost all cases it is better to not create an intermediate list, since it needs to be reserved and reallocated a bunch just to be consumed with another linq query again.  
  For all cases where the result is used multiple times however the enumerable should be collected first.  
  This is also quite dependent on the code that generates such a result. If that codereturns a reference to an existing list, we ofcourse just return that list. But if it generates a sequence we return the generator and don't create superfluous copies where it is not necessary.

- Changed several abstract output types to concrete ones, specifically changed `IReadOnlyList<T>` to just `List<T>`:
  This is relevant because in many cases functions were returning a readonly list that was created from a list, just to then be consumed by another function that would turn the ro list back into a normal list.
  This "turning back" requires a complete copy of the whole list, and just plainly wastes cpu cycles. I only did this in cases where the List is constructed as part of the first function call. I chose different approaches in cases where it is part of an internally stored cache and therefore "shared" between all consumers.

- Used structs instead of classes:  
  If there are lots of instances that get created of a specific type (e.g. Segment / Point) that are mostly readonly you want to have that be a struct. Structs don't need to be individually allocated but can live on the stack or in other datatypes, so it saves a lot of garbage collection and memory to have them be a struct.  
  Unfortunately there are som cases where a struct cannot be used, or is actually degrading performance. The biggest example are modifying list elements, as `list[3].value = 3` does not work if the generic type of the list is a struct type. The whole struct needs to be replaced.  
  For such use cases there unfortunately is no good solution in net standard except implementing a list type with ref indexing (which im not going to do).

- Implemented own sort algorithm (kinda):  
  I added a port of fluxsort to the repo, a stable sorting algorithm with restrictable and predictable memory usage. The implementation uses ArrayPooling where possible, but I went for speed over memory. The pooling is not ideal, but the best solution i could come up with for the port.

- Initialized lists to known or estimated lengths to prevent reallocations.

- Changed properties to fields wherever possible:
  This is very easy. Properties bad. There is no point in them existing. They add an additional layer of function calls to everything and refactoring code is trivially easy. If you don't _really_ need them, don't use them.
  the only reason to have them is if you want to use the binding system of Winforms or WPF, but those are limited and should not be the default.

### Recommendations and future work:

Code style / architecture:
- Move graph generation out of the parser. In general the architecture of  the library is very questionable. Generating a graph is not a parser task. 
- Clean up the comments: a lot of comments are just empty or pointless. Just remove empty ones, they are only noise when reading code.
- Leave empty lines! The `//` instead of empty lines breaks git diffing and merging. I had to do multiple 300 lines+ manual diffs where only a few liens changed during the final rebase, just because they were all glued together.
  I also don't quite understand where this even comes form, as empty liens are used for visual and mental separation of code blocks. Gluing the blocks together with empty comments not only introduces more noise, but it doesn't serve the original purpose of spacing out conceptually different segments since your eye glaze over the "spaces" because they are not empty.
- Rename `Abstract..` types to just the name. So instead of `AbstractDecoration` just call it `Decoration`. There is no value in this prefix. Same goes for `Generic...`.

Tests: 
- Add actual tests. At least one test of each encounter that gets special treatment, preferably additional ones for know edgecases. Each time a bug is reported and fixed add a regression test for that bug. 
- Add fuzztesting for the parser. This is a bit more complicated, but the evtc should easily be able to be simulated using something like https://github.com/nathaniel-bennett/winsharpfuzz.

Performance:
- Compress the html css and html output in general.
- Stream the html output.
- ~~Replace other small structures like Point3d with structs.~~


GW2EIEvtcParser/EIData/CombatReplay/Decorations/Forms/DoughnutDecoration.cs
```diff
@@ -1,74 +1,71 @@
-﻿using System;
-using System.Collections.Generic;
-using GW2EIEvtcParser.ParsedData;
+﻿using GW2EIEvtcParser.ParsedData;
 
-namespace GW2EIEvtcParser.EIData
+namespace GW2EIEvtcParser.EIData;
+
+public class DoughnutDecoration : FormDecoration
```
EliphasNUIT, line 5: Certain decorations became public, while others remain internal, what is the motivation behind that change?

SaculRennorb: That must be a leftover from testing, before I added the assembly attribute to share internals.
Ill address it shortly

SaculRennorb: On second thought, it might also have to do with serialization.
Using the more efficient compiletime-generated serializer requires explicit references to superclasses that may be encountered during serialization, and those superclasses as well as the base class must all share the same visibility

SaculRennorb: Yea so its the serialization.
Do you by chance have an example of decorations that are not public? I assume I forgot to add the explicit marker for those classes.

EliphasNUIT: Decorations/Images/BackgroundIconDecoration.cs is the only one remaining from what I see.
In that case, perhaps constructors of classes that became public should become internals?

EliphasNUIT: Wait, we only serialize the DecorationDescription classes though, not the Decoration classes.

GW2EIEvtcParser/EIData/CombatReplay/CombatReplay.cs
```diff
@@ -1,649 +1,667 @@
...
 
 
-        #region DEBUG EFFECTS
+    #region DEBUG EFFECTS
 
-        internal static void DebugEffects(AbstractSingleActor actor, ParsedEvtcLog log, CombatReplay replay, HashSet<long> knownEffectIDs, long start = long.MinValue, long end = long.MaxValue)
+    internal static void DebugEffects(AbstractSingleActor actor, ParsedEvtcLog log, CombatReplay replay, HashSet<long> knownEffectIDs, long start = long.MinValue, long end = long.MaxValue)
+    {
+        var effectEventsOnAgent = log.CombatData.GetEffectEventsByDst(actor.AgentItem)
+            .Where(x => !knownEffectIDs.Contains(x.EffectID) && x.Time >= start && x.Time <= end);
```
EliphasNUIT, line 187: The purpose of the unused variables in the debug methods it to have an easy access to certain values when breaking in that code, I'd very much prefer that they remain here, I use it to isolate effects and their GUIDs within certain windows.

GW2EIBuilders/HtmlModels/HtmlMetaData/SkillDto.cs
```diff
@@ -1,61 +1,93 @@
...
 using GW2EIEvtcParser;
 using GW2EIEvtcParser.EIData;
 using GW2EIEvtcParser.ParsedData;
 
-namespace GW2EIBuilders.HtmlModels.HTMLMetaData
+namespace GW2EIBuilders.HtmlModels.HTMLMetaData;
+
+
+[JsonConverter(typeof(SkillDataConverter))]
+public struct SkillData2
```
EliphasNUIT, line 11: That's not the best naming =D Maybe SerializableSkillDto?

GW2EIBuilders/RawFormatBuilder.cs
```diff
@@ -1,130 +1,71 @@
...
-        public void CreateXML(StreamWriter sw, bool indent)
-        {
-            CreateXML(_jsonLog, sw, indent);
-        }
+    /// <summary>
+    /// Writes the given JsonLog to the stream as json formatted text.
+    /// </summary>
+    public static void CreateJSON(JsonLog jsonLog, Stream stream, bool indent)
+    {
+        JsonSerializer.Serialize(stream, jsonLog, indent ? SerializerSettings.Indentent : SerializerSettings.Default);
```
EliphasNUIT, line 45: Does the serializer here also take into account the custom tuple converter?

SaculRennorb: It does not, as the generator needs separate instructions per root type.
I did indeed forget to add converters for the JsonLog.

GW2EIBuilders/HTMLBuilder.cs
```diff
@@ -1,308 +1,338 @@
...
+}
+
+// compile-time generated serialization logic
+[JsonSourceGenerationOptions(GenerationMode = JsonSourceGenerationMode.Metadata | JsonSourceGenerationMode.Serialization,
+    PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase,
+    IncludeFields = true, WriteIndented = false, NumberHandling = JsonNumberHandling.AllowNamedFloatingPointLiterals,
+    Converters = [ typeof(IntTupleArrayConverter) ]
+)]
+[JsonSerializable(typeof(LogDataDto))]
+[JsonSerializable(typeof(AgentConnector.AgentConnectorDescriptor))]
```
EliphasNUIT, line 36: I'm guessing we would not need those AgentConnector serializables if the GetConnected to API were to return a generic ConnectorDescriptor class with the others declared as JsonDerived? Just as a note to myself, trying to get used to System.Text.Json useage.

SaculRennorb: Yes. In the process of clearing up the visibility question im about to add the following comment to that code :D :

```cs
[JsonSerializable(typeof(LogDataDto))]
//NOTE(Rennorb): The following types need to be manually added to the generator, because they are used as opaque object fields.
// The generator cannot know the types of those fields, and we want to avoid having to use slow reflection to find out the members of those arbitrary types.
// Specifying them here tells the generator to generate switches for these specific types whenever an opaque Object needs to be serialized.
[JsonSerializable(typeof(AgentConnector.AgentConnectorDescriptor))]
[JsonSerializable(typeof(PositionConnector.PositionConnectorDescriptor))]
[JsonSerializable(typeof(AgentFacingConnector.AgentFacingConnectorDescriptor))]
...
```

SaculRennorb: I should also mention that this is not required to use the System.Text.Json serializer, only if you use the generated (de)serializers. These are quite a bit faster because they don't use reflection to find fields, but you need to specify more data in cases that use inheritance (which using "object" as the base is just a special case of).

GW2EIEvtcParser/EIData/CombatReplay/Decorations/Forms/DoughnutDecoration.cs
SaculRennorb: I think you are correct.
I removed the attributes and changed concrete decorations back to in internal visibility

GW2EIEvtcParser/EIData/CombatReplay/CombatReplay.cs
SaculRennorb: Added them back with a comment about this usage, and a define to toggle the functions of for production builds

GW2EIBuilders/HtmlModels/HtmlMetaData/SkillDto.cs
SaculRennorb: I'm all for good type names, and have a fair share of gripes towards the codebase around that myself, but that is not it chief. 

How about `SkillCastDto`  ?

GW2EIEvtcParser/EncounterLogic/Fractals/FractalLogic.cs
```diff
@@ -34,235 +34,231 @@ protected FractalLogic(int triggerID) : base(triggerID)
...
-            if (!log.AgentData.TryGetFirstAgentItem(target, out AgentItem agent))
-            {
-                return;
-            }
-            if (log.CombatData.TryGetEffectEventsByGUID(effectGUID, out IReadOnlyList<EffectEvent> effects))
+            return;
+        }
+        if (log.CombatData.TryGetEffectEventsByGUID(effect, out var events))
+        {
+            foreach (EffectEvent @event in events)
```
EliphasNUIT, line 240: I'm seeing such renaming patterns here and there, what is the reasoning?

GW2EIBuilders/RawFormatBuilder.cs
EliphasNUIT: Yeah, go for it.

GW2EIBuilders/HtmlModels/HtmlMetaData/SkillDto.cs
EliphasNUIT: Yeah, go for it.

GW2EIEvtcParser/EncounterLogic/Strikes/EndOfDragons/XunlaiJadeJunkyard.cs
```diff
@@ -12,453 +9,459 @@
...
+    {
+        base.ComputePlayerCombatReplayActors(p, log, replay);
+        if (p.GetBuffGraphs(log).TryGetValue(DeathsHandSpreadBuff, out BuffsGraphModel value))
         {
-            base.ComputePlayerCombatReplayActors(p, log, replay);
-            if (p.GetBuffGraphs(log).TryGetValue(DeathsHandSpreadBuff, out BuffsGraphModel value))
+            foreach (Segment segment in value.BuffChart)
             {
-                foreach (Segment segment in value.BuffChart)
+                //TODO(Rennorb) @correctnes: there was a null check here, i have no clue why.
```
EliphasNUIT, line 418: Most likely a very agressive defensive programming, I don't see how BuffChart could contain null values.

GW2EIEvtcParser/EncounterLogic/Fractals/FractalLogic.cs
SaculRennorb: `effectGUID` to `effect` ? The information about it being a GUID is now encoded in the type, so no need to have it in the variable name.

EliphasNUIT: I was rather curious about the "events" and "@event" renaming ^^ Sounded a little bit generic, I would have gone for "effectEvents" and "effectEvent" for better clarity

SaculRennorb: Ah - well, in some cases there where name collisions form the effectGUID(s) -> effects rename (since its now a distinct type) 
Then i realized that the old out param `effects` used in a bunch of places is straight up named incorrectly. It doesn't give you multiple effects, it gives you the events of that one specific effect.
`effectEvents` is certainly even more verbose, if you want me to change it to that we can do so np.

EliphasNUIT: @event makes it easy to track it down, it can be done later if you don't want to go through some files again.

SaculRennorb: to be fully honest im not sure i even changed all of them. I only changed the ones that were causing name collisions or where right under my nose. So if you wanted to clean that up it would need to be more methodical than just searching  the name.

GW2EIBuilders/RawFormatBuilder.cs
SaculRennorb: Added a new Tuple2ConverterFactory that replaces the old converter and added it to the JsonLog.

Quick question: why are there even two different json representations?

EliphasNUIT: One if officially documented and stable, the other one is unstable and is tailored in a way that matches the components of the html (it also does not contain data not exploited by the html, contrary to the documented JSON), so it changes anytime the components are moved around or reorganized.

GW2EIEvtcParser/ParsedData/CombatData.cs
```diff
@@ -1,1746 +1,1409 @@
...
         {
-            if (_breakbarDamageTakenData.TryGetValue(dst, out List<BreakbarDamageEvent> res))
+            effectEvents = GetEffectEventsByEffectID(effectGUIDEvent.ContentID);
+            if (effectEvents.Count > 0)
             {
-                return res;
+                return true;
             }
-            return new List<BreakbarDamageEvent>();
         }
```
EliphasNUIT, line 1033: Why the inconsistent behavior on the "out" of TryGets? Some put null when nothing is found, others a empty list.

GW2EIEvtcParser/ParsedData/CombatData.cs
```diff
@@ -1,1746 +1,1409 @@
...
-        }
+        effectEvents = null;
+        return false;
+    }
 
-        public IReadOnlyList<EffectEvent> GetEffectEventsByEffectID(long effectID)
+    /// <summary>
+    /// Appends effect events by the given agent and effect GUID.
+    /// </summary>
+    public void AppendEffectEventsBySrcWithGUID(AgentItem agent, GUID effect, List<EffectEvent> effectEvents)
```
EliphasNUIT, line 1077: Why are the Append methods public?

GW2EIEvtcParser/ParsedData/CombatData.cs
SaculRennorb: Assuming you are talking about the TryGetEffectEventsByGUID function; this did not change from the original implementation, which is here: https://github.com/baaron4/GW2-Elite-Insights-Parser/pull/933/files/8c9b50cc509cebc9413ebbdafee850160fe0c526#diff-d5e44689367f787448b7f0dfe2eee8beb2da1421aabd9d2e78d663e33736ba32L1306-L1320 

and also returns null on failure.

GW2EIEvtcParser/ParsedData/CombatData.cs
SaculRennorb: To be completely honest the reason is because i don't believe in the OOP paradigm of encapsulation and therefore declare almost everything public in my own work. I tried doing a reasonable job with this, but that one must have slipped through. We can certainly make it private if that is preferable.

GW2EIEvtcParser/ParsedData/CombatData.cs
EliphasNUIT: I don't have a strong opinion on wether it should be null or an empty list but I would prefer the behavior to be consistent. If your preference is "we have already created the list, just put it on the out", that is perfectly acceptable.

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItemBase.cs
```diff
@@ -1,132 +1,130 @@
...
+        else
         {
-            return GetSources();
+            distrib.Add(agent, new BuffDistributionItem(
+                cDur,
+                0, 0, 0, 0, 0));
         }
 
-        public override void SetBuffDistributionItem(BuffDistribution distribs, long start, long end, long buffID)
+        if (_buffStackItem.IsExtension)
```
EliphasNUIT, line 89: This won't work properly, StackID and SeedSrc are the only attributes that are constant during a BuffStackItem's lifecycle, for the same reason Start, Duration, Src and TotalDuration have to be stored, so does IsExtension.

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItemBase.cs
```diff
@@ -1,132 +1,130 @@
...
+    internal readonly BuffStackItem _buffStackItem;
+    internal readonly long _totalDuration;
+
+    protected internal BuffSimulationItemBase(BuffStackItem buffStackItem) : base(buffStackItem.Start, buffStackItem.Start + buffStackItem.Duration)
     {
-        private readonly AgentItem _src;
-        private readonly AgentItem _seedSrc;
-        private readonly bool _isExtension;
-        private long _totalDuration { get; }
+        _buffStackItem = buffStackItem;
```
EliphasNUIT, line 13: Is it really worth keeping the whole buff stack item alive for the whole duration of the processing of the log just to have a pointer on SeedSrc?

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItemBase.cs
```diff
@@ -1,132 +1,130 @@
...
-        }
+    public override long GetActualDuration()
+    {
+        return _totalDuration;
+    }
 
-        public override long GetActualDuration()
+    public override IEnumerable<AgentItem> GetSources()
+    {
+        return [ _buffStackItem.Src ];
```
EliphasNUIT, line 59: should return _src

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItem.cs
```diff
@@ -1,58 +1,58 @@
...
-namespace GW2EIEvtcParser.EIData.BuffSimulators
+namespace GW2EIEvtcParser.EIData.BuffSimulators;
+
+/// <summary> A segment of time with type <see cref="double"/> with inclusive start and inclusive end. </summary>
+using Segment = GenericSegment<double>;
+
+internal abstract class BuffSimulationItem : AbstractSimulationItem
 {
-    internal abstract class BuffSimulationItem : AbstractSimulationItem
+    //NOTE(Rennorb): I changed the element to have start + end instead of start + duration to fix a bug.
```
EliphasNUIT, line 11: What was the bug?

GW2EIEvtcParser/ParsedData/CombatData.cs
SaculRennorb: Ok, so after having a second look at this:
Why shouldn't the append methods be public? The TryGet methods that currently use them are are just wrappers around a loop of append calls, so why shouldn't they share the same visibility? Sometimes one or the other method might be more convenient to call, why take away that choice from the consumer and pretend to know which one will be more usefull?

This is also the main reason why I have such issues with encapsulation: It assumes you as the library author know better what the consumer wants than the consumer.

GW2EIEvtcParser/ParsedData/CombatData.cs
SaculRennorb: Yea but this is exactly why they behave differently in the first place: one of them creates the list no matter what, and the other one filters an enumerable to obtain the result. 

Maybe some renaming in in order here, because "success" doesn't even make sense in the context of merging multiple lists. I will look into just dropping the "Try" prefix from some of these and see if that cleans it up.

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItemBase.cs
SaculRennorb: There is some argument to be made for keeping it around, because it might reduce the amount of individual small deallocations, but this is a remnant from me trying to just keep the item around and not copy all the fields (see comments below).
I reverted it to store individual fields.

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItem.cs
SaculRennorb: To be fully honest don't quite remember. It had something to do with negative durations, but this section got chewed on quite a bit. Ive mainly kept it this way because the end time is the thing getting modified and read back for most operations, so it being represented by s+e makes it easier to reason about changes. 
As the comment states it should not matter much which way this is represented, but most operations are simpler in this version.

EliphasNUIT: I don't mind the change, end being the attribute matches better the useage indeed. As you have observed, it was initially following arc's start + duration logic. 
Without context, it seemed that you changed it to this in order to fix a bug that is currently present in master.

GW2EIEvtcParser/ParsedData/CombatData.cs
EliphasNUIT: That is quite the tangent, "To be completely honest the reason is because i don't believe in the OOP paradigm of encapsulation and therefore declare almost everything public in my own work" was a perfectly acceptable answer =D. The API does not expose something that is private nor allow something private to be modified. I saw that the method was only used locally, thus mu curiosity.

GW2EIEvtcParser/EIData/Buffs/BuffSimulators/BuffSimulationItems/BuffSimulationItem.cs
SaculRennorb: yea i get the concern. I don't think it is something on master, I'm confident it was something i introduced and subsequently fixed. I will look into it again before this gets finished.

GW2EIEvtcParser/ParsedData/CombatData.cs
SaculRennorb: I will have another look and find out if it would be sensible to use the append version in other places aswell, then report back before we resolve this.

GW2EIEvtcParser/EIData/Statistics/StatisticsHelper.cs
```diff
@@ -1,265 +1,276 @@
...
-        private readonly List<Buff> _presentGearbuffs = new List<Buff>();//Used only for Gear Buff tables
-        private readonly List<Buff> _presentNourishments = new List<Buff>();
-        private readonly List<Buff> _presentEnhancements = new List<Buff>();
-        private readonly List<Buff> _presentOtherConsumables = new List<Buff>();
-        private readonly Dictionary<Player, HashSet<Buff>> _presentRemainingBuffsPerPlayer = new Dictionary<Player, HashSet<Buff>>();
+    //Positions for group
+    private List<ParametricPoint3D?>? _stackCenterPositions = null;
+    private List<ParametricPoint3D?>? _stackCommanderPositions = null;
 
+    //TODO(Rennorb) @cleanup: So ... why are these parametric points, but we still add null values between them?
```
EliphasNUIT, line 161: Convenience mainly. It guarentees the same delta between elements and easily find start offsets. 
In GetDistanceToTarget, the number of skipped elements on processed actor gives us immediately the start index on the stackCenter/commanderPosition lists.

GW2EIEvtcParser/EIData/Actors/AbstractSingleActor.cs
```diff
@@ -1,1143 +1,915 @@
...
-        {
-
-            if (CombatReplay == null)
-            {
-                InitCombatReplay(log);
-            }
-            return CombatReplay.Decorations.GetCombatReplayRenderableDescriptions(map, log, usedSkills, usedBuffs);
-        }
-        protected virtual void InitAdditionalCombatReplayData(ParsedEvtcLog log)
+    //TODO(Rennorb) @cleanup: There is an argument to be made here that in all the places where this is used you probably want to add a decoration either way, regardless if this fails or not.
```
EliphasNUIT, line 509: Could you explain what do you mean by this? Places that use TryGetCurrentFacingDirection are making use of the vector to give a proper orientation to the decoration.

SaculRennorb: So this functions gives you the facing direction of its actor - but it can fail. 
So what do we do when this fails? Iirc in most cases we don't do the thing we originally wanted to do, so for example we don't create a decoration. but this is wired; my argument is that we probably want to create that decoration either way, and maybe just give it the wrong facing direction. Just because we don't have a direction doesn't mean we don't have a position.

The todo is on the getter because this is the centralized place from which you can find all the usages. It was easier than investigating every single one when i thought about that potential "problem".

EliphasNUIT: Why would we want to show a decoration that is facing the wrong way though? ^^ A rectangle or a pie facing the wrong way completely changes the meaning of the mechanic. I don't believe we use Facing directions for decorations that are invariant by rotation.

SaculRennorb: Yea i also figured that exact issues with my original proposal, so maybe this is not the best solution. But not showing anything at all is even worse imo. Thats the reason why there is a not about wanting to look into this more.

EliphasNUIT:
Thanks!

EliphasNUIT:
To be investigated: ArrayPool seems to allocate too much memory, memory useage increases up to a point while re-parsing a single file.

GW2EIEvtcParser/ParsedData/CombatData.cs
SaculRennorb: Alright, looks like its not really feasible to propagate the append methods even more, but they might come in handy at some point down the line.

PR #934, opened 2024-11-06
Fixed some issues with Harvest Temple combat replay and cleanup

Linkaaaaa:
- Fixed Void Pools in Challenge Mode lasting 300000ms instead of 240000ms (5 minutes instead of 4).
- Fixed Void Pools in Normal Mode lasting 25000ms instead of 24000ms.
- Appropriately using TryGetEffectEventsByDstWithGUID for Void Pool and Spread selections.
- Fixed Greens not despawning when transitioning to another dragon (except Soo Won, they don't despawn there).
- Removed unnecessary DeadEvent and DespawnEvent checks on the spreads. A dead player with the spread will damage another player in the AoE.
- Fixed Spreads lasting 5500ms instead of 5000ms in Challenge Mode.
- Using Dynamic End Event for effects on player for logs that have it as of EVTC20241030.
- General cleanup and documentation.


GW2EIEvtcParser/EncounterLogic/Strikes/EndOfDragons/HarvestTemple.cs
```diff
@@ -1700,138 +1698,138 @@ private AbstractSingleActor FindActiveOrNextDragonVoid(long time)
...
         {
-            var redSelectedEffectsOnPlayer = redSelectedEffects.Where(x => x.Dst == p.AgentItem).ToList();
-            foreach (EffectEvent redSelectedEffect in redSelectedEffectsOnPlayer)
+            foreach (EffectEvent effect in effects)
             {
-                int duration = 7000;
-                int start = (int)redSelectedEffect.Time;
-                AbstractSingleActor dragonVoid = FindActiveOrNextDragonVoid(redSelectedEffect.Time);
+                long duration = 7000;
+                (long start, long end) lifespan = effect.HasDynamicEndTime ? effect.ComputeDynamicLifespan(log, 7936) : (effect.Time, effect.Time + duration);
```
EliphasNUIT, line 1735: why the duration given to compute dynamic is different than the declared duration? Is the declared one wrong?

Linkaaaaa: 7936 works correctly with dynamic end in normal mode, in challenge mode it doesn't and the aoe lingers on the player for those extra 936 milliseconds (in game you can see it too), so i'm using the fixed 7000 in that case

EliphasNUIT:
Thanks!

PR #945, opened 2024-11-22
Update skills API cache (add W8 skills)

stellarashes:
Update skills API cache, mainly to add W8 skills data


stellarashes:
hmm looks like i screwed up something and can't build with this, will fix it first

PR #947, opened 2024-11-23
Combat replay maps for Wing 8

LunusLupus:
- Added first draft of CR maps for all three W8 bosses
  - Only basic outline for now, more details for Ura and Greer might be helpful
- Positioning is preliminary only


EliphasNUIT:
Thanks!

PR #953, opened 2024-12-04
Fix max duration effects decorations not being displayed due to int cast

Linkaaaaa:
Also set Healing Spring to 300000 which is the timeout for traps in game


EliphasNUIT:
Thanks!

PR #957, opened 2024-12-12
Fixed some decoration sizes and adjusted miasma for the new patch

Linkaaaaa:
Sweep the Mold, Rake the Rot and Stomp the Growth have 3 different radiuses based on the skill id casted.
Sweep the Mold also has a Y 40 offset which Rake the Rot does not have, so I've manually added the value to match the in-game indicator.
Fixed the Miasma clouds not showing up in the new patch, they use a new effect, also adjusted the radius approximation.
Added Eruption of Rot buff, it's applied when the green ends to every player that stood in it, probably used to down a player with 2+ stacks, and added mechanics checks with it.
Renamed some things and added some comments for better clarity,
Fixed the Blob of Blight indicator lasting for the entire encounter if the npc is killed before the blob spawns.


EliphasNUIT:
Make sure to limit none relevant mechanics for current build via "WithBuilds" API.

Linkaaaaa:
I have added the build checks to the relevant mechanics, though I've removed it later for the Decima's death mechanics, they can still activate if you are in downstate in the hitbox

Linkaaaaa:
Also got reported that Rising Pressure was only checking against strike damage, it was incorrect for only 2 of the 3 modifiers we're checking

EliphasNUIT:
Thanks!

PR #960, opened 2025-01-10
Image migration

EliphasNUIT:
Refactoring + migrate to assets.gw2dat (from wiki.guildwars2) when easily possible.


EliphasNUIT:
General restructuration is finished, migration will continue on another PR.

PR #962, opened 2025-01-12
Improve fight detection based on target agent type

Zerthox:
Current fight detection does not account for potential collisions between NPC species ids and gadget pseudo ids. I've been made aware of at least one such collision existing: the Eternal Flame gadget boss in Citadel of Flame is incorrectly identified as golem in EI.
This PR changes the fight detection to determine the target agent type first and then check possible targets for that type in order to handle collisions correctly.


EliphasNUIT:
Looks good, thanks!

PR #963, opened 2025-01-14
Fix Sabetha phases & targets

Zerthox:
This fixes incorrect mini boss phases on Sabetha and their targets. The "latest" mini boss is prioritized as main target and any other mini bosses are added as secondary targets.

It also changes behavior of adding phase targets to help prevent duplicate primary & secondary targets:
- Adding a new primary target removes the same actor from the list of secondary targets (if present)
- Adding a new secondary target is skipped if the same actor is already present as primary target


EliphasNUIT:
Thanks!

PR #964, opened 2025-01-19
Improve simple rotation

Zerthox:
- Clicking a cast highlights all casts of the same skill
- Tooltip shows cast counts for the hovered skill & overall for all skill casts


EliphasNUIT:
Thanks!

PR #965, opened 2025-01-22
Fix some autoattacks and Necromancer Shroud labeling

Linkaaaaa:
Gunsaber autoattack chain and Firebrand Chapter 1 skills are now categorised as autoattacks.
Necromancer shroud is now categorised as a weapon swap.
Separated Reaper Shroud and Harbinger Shroud from IsShroudTransform, Reaper Shroud is already a weapon swap.
Added Path of Fire release build number.

### DIFF
--- a/GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/NecromancerHelper.cs
+++ b/GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/NecromancerHelper.cs
@@ -123,9 +123,7 @@ internal static class NecromancerHelper
 
     public static bool IsShroudTransform(long id)
     {
-        return _shroudTransform.Contains(id)
-            || ReaperHelper.IsReaperShroudTransform(id)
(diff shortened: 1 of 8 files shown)


GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/NecromancerHelper.cs
```diff
@@ -123,9 +123,7 @@ internal static class NecromancerHelper
 
     public static bool IsShroudTransform(long id)
     {
-        return _shroudTransform.Contains(id)
-            || ReaperHelper.IsReaperShroudTransform(id)
-            || HarbingerHelper.IsHarbingerShroudTransform(id);
+        return _shroudTransform.Contains(id);
```
EliphasNUIT, line 126: should be restored, add a public static bool IsDeathShroudTransform for your weapon swap instead

EliphasNUIT:
Thanks!

PR #968, opened 2025-01-30
Fix extra triggers of relic of the stormsinger

Linkaaaaa:
According to my tests, this should be working correctly now

### DIFF
--- a/GW2EIEvtcParser/EIData/ProfHelpers/ProfHelper.cs
+++ b/GW2EIEvtcParser/EIData/ProfHelpers/ProfHelper.cs
@@ -153,7 +153,7 @@ internal static class ProfHelper
         new EffectCastFinder(RelicOfTheStormsingerChain, EffectGUIDs.RelicOfTheStormsinger)
             .UsingChecker((effectEvent, combatData, agentData, skillData) => 
             {
-                return combatData.GetBuffData(RelicOfTheStormsingerBuff).Where(x => x is BuffRemoveAllEvent && Math.Abs(x.Time - effectEvent.Time) < ServerDelayConstant).FirstOrDefault() != null;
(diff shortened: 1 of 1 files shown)


GW2EIEvtcParser/EIData/ProfHelpers/ProfHelper.cs
```diff
@@ -150,9 +150,11 @@ internal static class ProfHelper
             .UsingOrigin(InstantCastFinder.InstantCastOrigin.Gear),
         new EffectCastFinder(RelicOfSorrowBuff, EffectGUIDs.RelicOfSorrow3)
             .UsingOrigin(InstantCastFinder.InstantCastOrigin.Gear),
-        // If this collides with something else we could check via buff remove with the condition dst == src
         new EffectCastFinder(RelicOfTheStormsingerChain, EffectGUIDs.RelicOfTheStormsinger)
-            .UsingSecondaryEffectChecker(EffectGUIDs.RelicOfTheStormsinger) // Effect triggers at least twice
+            .UsingChecker((effectEvent, combatData, agentData, skillData) => 
+            {
+                return combatData.GetBuffData(RelicOfTheStormsingerBuff).Where(x => x is BuffRemoveAllEvent && Math.Abs(x.Time - effectEvent.Time) < ServerDelayConstant).FirstOrDefault() != null;
```
EliphasNUIT, line 156: combatData.HasLostBuff API can be used here instead.
Perhaps we should leave the double effect check also to be safe?

Linkaaaaa: The incorrect trigger on the htcm log was caused by 4 effects triggering within roughly 2 seconds, so i've removed it entirely

EliphasNUIT:
Thanks!

PR #969, opened 2025-02-04
Added new deploy siege IDs and moved WvW icons

Linkaaaaa:
Added the new infinite siege skill ids and some leftover skills for mortar and burning oil
Moved all of the WvW icons to SkillImages, used gw2dat and render.guildwars2.com where available
Reordered some stuff around and fixed indentation of the wvw region


EliphasNUIT:
Thanks!

PR #978, opened 2025-02-23
Separate Cerus Embodiment targets into killed & not killed

Zerthox:
This adds killed Embodiments to the main phase and changes the priority of not killed Embodiments to non blocking in the split phases (as long as one of the Embodiments in that phase was killed).


EliphasNUIT:
Thanks!

PR #981, opened 2025-02-27
More instant cast finders

Linkaaaaa:
Elementalist
- Healing Ripple (PvE & WvW)
- Flow Like Water

Ranger
- Evasive Purity

Druid
- Blood Moon

Untamed
- Restorative Strikes

Necromander
- Spiteful Renewal

Harbinger
- Implacable Foe

Fixed Implacable Foe having the incorrect ID
Added Restorative Strikes buff
Adjusted Implacable Foe to be strike damage only

If some of these aren't worth having, let me know. I was considering adding Fresh Air too since it's similar to Healing Ripple.

### DIFF
--- a/GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/HarbingerHelper.cs
+++ b/GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/HarbingerHelper.cs
@@ -18,7 +18,6 @@ internal static class HarbingerHelper
     [
         new BuffGainCastFinder(EnterHarbingerShroud, HarbingerShroud).UsingBeforeWeaponSwap(true),
         new BuffLossCastFinder(ExitHarbingerShroud, HarbingerShroud).UsingBeforeWeaponSwap(true),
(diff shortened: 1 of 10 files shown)


GW2EIEvtcParser/EIData/ProfHelpers/Necromancer/HarbingerHelper.cs
```diff
@@ -18,6 +18,7 @@ internal static class HarbingerHelper
     [
         new BuffGainCastFinder(EnterHarbingerShroud, HarbingerShroud).UsingBeforeWeaponSwap(true),
         new BuffLossCastFinder(ExitHarbingerShroud, HarbingerShroud).UsingBeforeWeaponSwap(true),
+        new BuffGainCastFinder(ImplacableFoe, ImplacableFoe).UsingOrigin(EIData.InstantCastFinder.InstantCastOrigin.Trait),
```
EliphasNUIT, line 21: That's going to proc unconditionaly on each harbinger shroud enter, I don't feel like it warrants an instant cast detection.

EliphasNUIT:
Thanks! No need for fresh air, adding "gaining buff coming from a single skill, that is already present in rotation and with no condition attached" would simply bloat the rotation tab.

PR #983, opened 2025-03-03
ArcDPSEnums cleanup

Linkaaaaa:
- Moved TargetIDs, TrashIDs, MinionIDs and ChestIDs to the class SpeciesIDs
- Moved GW2Builds to its own class file
- Added documentation for each build number with a patch note wiki link
- Sorted and cleaned up usings in various files
- Added "using static GW2EIEvtcParser.ArcDPSEnums;" to all the files that make use of it and removed any unnecessary ArcDPSEnums.XXX reference


EliphasNUIT:
Thanks!

PR #985, opened 2025-03-06
Add Envious Gaze boonstrip mechanic and small fix

Linkaaaaa:
The community has requested a mechanic to display who has been boonstrip by Envious Gaze while reviewing the Combat Replay. The current issue is that, even though you can jump the wall, it only jumps the boonstrip but not the damage, meaning the Envious Gaze hit mechanic will trigger and it doesn't tell if the player has lost boons.

The result is this:

The change in the decoration is to prevent a fake cast while Envy is despawning, this case happens only while phasing to 10% while the Embodiment is already casting.

Video: https://www.youtube.com/watch?v=9MF1RL-v9Gg
Log of the phasing in the video: [20250304-150235.zip](https://github.com/user-attachments/files/19115774/20250304-150235.zip)
Link of how it looks currently: https://dps.report/Dwwu-20250304-150235_cerus

Added IsUnknown, IsReduced, IsInterrupted, IsFull and IsInstant APIs to CastEvent, in use for this Cerus fix and updated other files accordingly.


EliphasNUIT:
Thanks!

PR #1000, opened 2025-04-05
Combat Replay skill casts cleanup

Linkaaaaa:
# Scope:

- Remove multiple iterations of the IEnumerable returned by GetCastEvents
- - If multiple iterations are necessary due to how the Combat Replay works, cast to ToList(), eg Dhuum, Dagda, Cerus
- Remove unnecessary iterations of Instant Casts by using the new API GetAnimatedCastEvents
- - This mostly affects raid encounters with aura damage. For a random Vale Guardian log, out of 92 casts, 50 were aura damage, now only 42 will be iterated.

The main part of the rework is the introduction of a switch to iterate GetAnimatedCastEvents only once.

```C#
foreach (CastEvent cast in target.GetAnimatedCastEvents(log, log.FightData.FightStart, log.FightData.FightEnd))
{
    switch (cast.SkillId)
    {
        case SkillId1:
            // Decoration Creation
            break;
        case SkillId2:
            // Decoration Creation
            break;
        default:
            break;
    }
}
```

## Other:

- Cleaned up Combat Replays overall, compacted code and cleaned up a lot of unnecessary (int)cast.Time that were leftover from when decorations accepted int instead of long.
- Transitioned some colors from string to the Colors class

### DIFF
--- a/GW2EIEvtcParser/EIData/Actors/Minions.cs
+++ b/GW2EIEvtcParser/EIData/Actors/Minions.cs
@@ -271,6 +271,21 @@ public class Minions : Actor
         }
         return CastEvents.Where(x => KeepIntersectingCastLog(x, start, end));
     }
+
+    public override IEnumerable<AnimatedCastEvent> GetAnimatedCastEvents(ParsedEvtcLog log, long start, long end)
+    {
+        return log.CombatData.GetAnimatedCastData(AgentItem).Where(x => x.Time >= start && x.Time <= end);
+    }
+
+    public override IEnumerable<InstantCastEvent> GetInstantCastEvents(ParsedEvtcLog log, long start, long end)
+    {
+        return log.CombatData.GetInstantCastData(AgentItem).Where(x => x.Time >= start && x.Time <= end);
+    }
+
+    public override IEnumerable<WeaponSwapEvent> GetWeaponSwapEvents(ParsedEvtcLog log, long start, long end)
+    {
+        return log.CombatData.GetWeaponSwapData(AgentItem).Where(x => x.Time >= start && x.Time <= end);
+    }
     #endregion CAST
 
     internal bool IsActive(ParsedEvtcLog log)
--- a/GW2EIEvtcParser/EIData/Actors/Minions.cs
+++ b/GW2EIEvtcParser/EIData/Actors/Minions.cs
@@ -271,21 +271,6 @@ public class Minions : Actor
         }
         return CastEvents.Where(x => KeepIntersectingCastLog(x, start, end));
     }
-
-    public override IEnumerable<AnimatedCastEvent> GetAnimatedCastEvents(ParsedEvtcLog log, long start, long end)
-    {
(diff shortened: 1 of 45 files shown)


GW2EIEvtcParser/EIData/Actors/SingleActor.cs
```diff
@@ -519,12 +519,28 @@ public override IEnumerable<CastEvent> GetIntersectingCastEvents(ParsedEvtcLog l
...
+    {
+        return log.CombatData.GetAnimatedCastData(AgentItem).Where(x => x.Time >= start && x.Time <= end);
+    }
+
+    public override IEnumerable<InstantCastEvent> GetInstantCastEvents(ParsedEvtcLog log, long start, long end)
+    {
+        return log.CombatData.GetInstantCastData(AgentItem).Where(x => x.Time >= start && x.Time <= end);
+    }
+
+    public override IEnumerable<WeaponSwapEvent> GetWeaponSwapEvents(ParsedEvtcLog log, long start, long end)
```
EliphasNUIT, line 533: This will not be equivalent to iterating over weapon swaps from GetCastEvents out put, if not necessary, it is better to not have that API.

GW2EIEvtcParser/EIData/Actors/Minions.cs
```diff
@@ -271,6 +271,21 @@ public override IEnumerable<CastEvent> GetIntersectingCastEvents(ParsedEvtcLog l
         }
         return CastEvents.Where(x => KeepIntersectingCastLog(x, start, end));
     }
+
+    public override IEnumerable<AnimatedCastEvent> GetAnimatedCastEvents(ParsedEvtcLog log, long start, long end)
```
EliphasNUIT, line 275: That implement does not make sense, minions are a group of multiple AgentItems, those will return events for only a single agent within the pack.
Also that would require events to be sorted by time at each call, which is going to be very inneficient. I think you can limit those APIs to SingleActor.

GW2EIEvtcParser/EncounterLogic/Strikes/SotO/CosmicObservatory.cs
```diff
@@ -82,78 +82,76 @@ protected override CombatReplayMap GetCombatMapInternal(ParsedEvtcLog log)
...
+                            uint length = 1500;
+                            uint width = 100;
+                            int castDuration = 6700;
+                            lifespan = (cast.Time, cast.Time + castDuration);
+
+                            // The mechanic gets cancelled during the intermission phases since the CM release.
+                            // Before then, the mechanic would continue during the phase and shoot.
+                            if (log.LogData.GW2Build >= GW2Builds.DagdaNMHPChangedAndCMRelease)
+                            {
+                                foreach (CastEvent demonicBlastCast in casts.Where(x => x.SkillId == DemonicBlast))
```
EliphasNUIT, line 134: better to store the result of the Where somewhere, instead of doing it for each instance of ShootingStarrs

EliphasNUIT:
Looks good, you can do the mentionned changes and it is good to go.

EliphasNUIT:
Thanks!

PR #1004, opened 2025-04-12
Relics update

Linkaaaaa:
Added Relic of the Steamshrieker
Fixed Relic of the Blightbringer using the incorrect icon
Moved relics icons from wiki.guildwars2 to render.guildwars2


EliphasNUIT:
Thanks!

PR #1011, opened 2025-04-25
Professions cleanup 3

Linkaaaaa:
Added Illusionary Membrane new buff simulation and related damage modifier
Updated Jaunt to be using effect events and locked old damage finder
Added icon for Debilitating Arc roll missing since arcdps can detect movement skills
Added older version of Hardened Armor
Documented ArcDPS skillIDs


EliphasNUIT:
Thanks!

PR #1014, opened 2025-04-30
Move trigonometry methods to their own class in MathUtils

Linkaaaaa:
Moved the Radian/Degree methods from ParserHelper to its own class in the MathUtils folder.
Looks like some auto formatting leaked in for KC but it's fine.


EliphasNUIT:
Thanks

PR #1018, opened 2025-05-13
Fix C# Dev Kit error on Linux

Linkaaaaa:
The VSC C# Dev Kit errors out on GW2EIParser.csproj because of this tag not being set.


EliphasNUIT:
Thanks.

PR #1020, opened 2025-05-21
Update Malicious Intent

Linkaaaaa:
Back when Cerus CM released, I couldn't find a buff applied to players targeted by malices, today I found it and it works retroactively to the very first log i have of the CM.
This cuts down a lot of extra steps that were taken to make the tethers work.


EliphasNUIT:
Thanks!

PR #1021, opened 2025-05-23
Projectile Decorations 2

Linkaaaaa:
Draft

# Dhuum
- Added Enforcer, Messenger, Spider and generic collection orbs
- Added Deathling icon

# API
- Added GenericSegment with tuple


Linkaaaaa:
@EliphasNUIT Ready for review again before it gets too large, there are more things I want to add but this is enough for now

PR #1032, opened 2025-06-14
Breakbar improvments

Zerthox:
This fixes the breakbar damage values from generic breakbar damage events, which combines regeneration + soft CC. Positive values are healing and negative values are damage.

It also adds the total breakbar damage dealt to the tooltip of breakbar phases.


EliphasNUIT:
Thanks!

PR #1038, opened 2025-06-29
Adjust Undying Light post patch

Linkaaaaa:
The achievement behaves a bit differently than the others we've previously implemented


EliphasNUIT:
Thanks!

PR #1043, opened 2025-07-16
Fix Kinfall phases & add Falling Ice

Zerthox:
This fixes Kinfall phase detection for the recent patch & adds replay decorations for Falling Ice.


EliphasNUIT:
Thanks!

PR #1047, opened 2025-07-21
Added Flame Expulsion

Selebrator:
Add [Flame Expulsion](https://wiki.guildwars2.com/wiki/Flame_Expulsion) of [Pyromancer's Puissance](https://wiki.guildwars2.com/wiki/Pyromancer%27s_Puissance) to rotation views.


GW2EIEvtcParser/EIData/ProfHelpers/Elementalist/ElementalistHelper.cs
```diff
@@ -36,6 +36,8 @@ internal static class ElementalistHelper
             .UsingDisableWithEffectData(),
         new DamageCastFinder(Sunspot, Sunspot)
             .UsingOrigin(EIData.InstantCastFinder.InstantCastOrigin.Unconditional),
+        new DamageCastFinder(FlameExpulsion, FlameExpulsion)
```
EliphasNUIT, line 39: "After a short delay, release a torrent of flame"
Does this behave like an instant damage event? Sounds like there is some sort of build-up. Does the release behave like a projectile? A shockwave?
Ultimately the question is, does the time at which damage is inflicted matches the time at which Flame Expulsion starts?

PR #1061, opened 2025-08-13
Fix envious gaze decoration, ura phases and some decima adjustments

Linkaaaaa:
Fix Ura phases not working correctly when the first HP update is 0.0%.
Fix Envious Gaze when phasing into 10% while the indicator is present.
Fix Envious Gaze back beam despawning early when ending 80% and 50% split phases, not matching in game behavior.
General small cleanup for Cerus.
Updated Earthrend, Seismic Crash and Seismic Reposition decorations
Updated Decima mechanics


GW2EIEvtcParser/LogLogic/Raids/W8/Bosses/DecimaTheStormsinger.cs
```diff
@@ -830,6 +829,35 @@ private static void AddThunderAoE(SingleActor actor, ParsedEvtcLog log, CombatRe
...
+    /// <summary>
+    /// Red death zone underneath Decima during Earthrend, Seismic Crash and Seismic Reposition.
+    /// </summary>
+    /// <param name="time">Time at which the other mechanic effect plays.</param>
+    /// <param name="isEarthrend">Wether the mechanic is Earthrend of a Seismic.</param>
+    private static void AddDeathZone(NPC target, ParsedEvtcLog log, CombatReplay replay, long time, bool isEarthrend)
+    {
+        (long start, long end) lifespan;
+
+        if (log.CombatData.TryGetEffectEventsBySrcWithGUID(target.AgentItem, EffectGUIDs.DecimaJumpAoEUnderneath, out var deathZone))
```
EliphasNUIT, line 841: This looks not very efficient, would be better to create a ComputeLifespanWithSecondaryEffectsAndPosition that takes an array of GUIDs that use TryGetEffectEventsBySrcWithGUIDs and have only a single loop instead of two

GW2EIEvtcParser/LogLogic/Raids/W8/Bosses/DecimaTheStormsinger.cs
```diff
@@ -830,6 +829,35 @@ private static void AddThunderAoE(SingleActor actor, ParsedEvtcLog log, CombatRe
...
+        {
+            EffectEvent effect = deathZone.FirstOrDefault(x => Math.Abs(x.Time - time) < ServerDelayConstant)!;
```
EliphasNUIT, line 843: That ! is not necessary, either use "var effect" or "EffectEvent? effect".

EliphasNUIT:
Thanks!

PR #1070, opened 2025-08-30
Removed raidar mentions

Linkaaaaa:
Kept in the option to have multiple urls in case wingman ever adds one in the response


EliphasNUIT:
Thanks.

PR #1075, opened 2025-09-07
Automatic updater

Linkaaaaa:
WIP, there's a lot i still want to rework so don't really need to review it yet


GW2EIParser/MainForm.Designer.cs
```diff
@@ -34,343 +34,356 @@ protected override void Dispose(bool disposing)
...
+        OperatorBindingSource = new BindingSource(components);
+        TlpMainWindow = new ToolTip(components);
+        BtnClearFailed = new Button();
+        BtnDiscordBatch = new Button();
+        LogFileWatcher = new FileSystemWatcher();
+        LblVersion = new Label();
+        LblWatchingDir = new Label();
+        ChkApplicationTraces = new CheckBox();
+        ChkAutoDiscordBatch = new CheckBox();
+        BtnCheckUpdates = new Button();
```
EliphasNUIT, line 62: The button should be disabled while the parser is running (QueueOrRunOperation and _RunNextOperation).

GW2EIParser/MainForm.cs
```diff
@@ -877,4 +878,24 @@ private void EnableSettingsWatcher(object sender, EventArgs e)
...
+        var mainForm = this;
+        var uiScheduler = TaskScheduler.FromCurrentSynchronizationContext();
+        Task.Factory.StartNew(async () =>
+        {
+            if (await updater.NewReleaseCheckerAsync())
+            {
+                mainForm.Invoke(() => // Must run on UI thread
+                {
+                    var updaterForm = new UpdaterForm(updater);
+                    updaterForm.Show();
```
EliphasNUIT, line 896: ShowDialog(mainForm) to disable interactions on the main form while the update window is open.

GW2EIUpdater/Updater.cs
```diff
@@ -0,0 +1,177 @@
+﻿using System.Diagnostics;
+using System.IO.Compression;
+using System.Reflection;
+using Newtonsoft.Json;
+
+namespace GW2EIUpdater;
+
+public class Updater
```
SaculRennorb, line 8: Why is this a class at all? You don't need multiple instances. This could just be `static class`. 
I personally would prefer just having a static method `UpdateInfo CheckForUpdate()`, then a second one `void DownloadAndUpdate(localFileName, UpdateInfo, remoteFileName)` that does the updating, where `UpdateInfo` is a struct of the versioning and download information, with a `bool IsUpadteAvialable()` either on the `UpdateInfo` or on the static class (and taking in an `UpdateInfo`).  

If you want to keep this an instantaible class, at least make use of oop features:
- make this class abstract
- implement a concrete version in UI / CLI that has the concrete information for that specific executable, e.g. only specify names and paths in the concrete version. That way you don't need this `downloadCli` variable that you pass on instantiation. e.g. make this have `abstract string DownloadName { get; }` (and similar) that then gets overridden by the hypothetical `CliUpdater` / `UIUpdater` that then set concrete values for their specific case.

GW2EIUpdater/Updater.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        string downloadUrl = string.Empty;
+        
+        // Windows: C:\Users\User\AppData\Local\Temp\
+        // Linux: /tmp/
+        string tempPath = Path.GetTempPath();
+
+        // Windows: C:\Users\User\AppData\Local\Temp\GW2EIUpdateTemp\
+        // Linux: /tmp/GW2EIUpdateTemp/
+        string folderPath = Path.Combine(tempPath, EI_TempFolder);
```
SaculRennorb, line 102: This should not be required. The temp path is by definition already a temporary folder for you to throw files in.
If you want to avoid collisions then there is [GetTempFileName](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=net-9.0)

GW2EIUpdater/Updater.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        // Create folder if it doesn't exist
+        Directory.CreateDirectory(folderPath);
```
SaculRennorb, line 105: This also becomes unnecessary if you remove the additional directory for temp files.

GW2EIUpdater/Updater.cs
```diff
@@ -0,0 +1,177 @@
...
+            Process.Start(new ProcessStartInfo
+            {
+                UseShellExecute = true,
+                FileName = $"{folderPath}\\{EIUpdater_Executable}",
+                Arguments = $"{AppContext.BaseDirectory} {Executable}",
+            });
+        }
+        catch
+        {
+            throw new DllNotFoundException("Executable not found");
```
SaculRennorb, line 174: You know exactly what failed here, why give such a vague error message?

GW2EIParserCLI/Program.cs
```diff
@@ -17,6 +17,17 @@ private static int Main(string[] args)
...
+        try
+        {
+            // TODO restore this comment before merging
+            // Directory.Delete(Path.Combine(Path.GetTempPath(), Updater.EI_TempFolder), true);
+        }
+        catch
+        {
+
+        }
+
```
SaculRennorb, line 30: The temp folder does not necessarily need cleanup, but I also don't see any reason as to why you couldn't do this in the updater after you have started the updated version.

GW2EIParser/Program.cs
```diff
@@ -54,6 +55,17 @@ private static int Main(string[] args)
...
+        try
+        {
+            // TODO restore this comment before merging
+            // Directory.Delete(Path.Combine(Path.GetTempPath(), Updater.EI_TempFolder), true);
+        }
+        catch
+        {
+
+        }
+
```
SaculRennorb, line 68: Same as with other other copy of this in CLI

GW2EIParser/Properties/AssemblyInfo.cs
```diff
@@ -12,8 +12,8 @@
 
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 #if !DEBUG
-[assembly: AssemblyVersion("3.13.0.0")]
+[assembly: AssemblyVersion("3.12.0.0")]
```
EliphasNUIT, line 15: Don't forget to rollback this.

GW2EIParser/MainForm.cs
```diff
@@ -877,4 +892,26 @@ private void EnableSettingsWatcher(object sender, EventArgs e)
...
     }
+
+    private void BtnCheckUpdates_Click(object sender, EventArgs e)
+    {
+        AddTraceMessage("Updater: Checking for updates");
+
+        Task.Factory.StartNew(async () =>
+        {
+            Updater.UpdateInfo info = await Updater.CheckForUpdate("GW2EI.zip");
+            if (info.UpdateAvailable)
```
EliphasNUIT, line 903: The if and else should update UpdateAvailable/UpdateLastChecked settings and refresh the label text

GW2EIParserCommons/App.config
```diff
@@ -0,0 +1,132 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
EliphasNUIT, line 1: What's this file? Auto generated?

GW2EIUpdater/Updater.cs
```diff
@@ -0,0 +1,180 @@
...
+        public bool UpdateAvailable = update;
+    }
+
+    /// <summary>
+    /// Elite Insights temporary folder name in the system 'Temp' folder.<br></br>
+    /// 'GW2EIUpdateTemp' for the UI and 'GW2EICLIUpdateTemp' for the CLI.
+    /// </summary>
+    public static string TempFolderName { get; private set; } = string.Empty;
+    
+    private static readonly bool _downloadCLI = false;
```
EliphasNUIT, line 47: This does not make much sense, it should be an argument for DownloadAndUpdate method or deduced from file name like in CheckForUpdate

Linkaaaaa: Yes I forgot to remove this one when i introduced isCli through file name

PR #1139, opened 2026-04-19
Added target stats % spread for some stats

Linkaaaaa:
Power vs Condition % and Healing Power vs Conversion %

From here maybe check if there is a good way to showcase total healing vs total barrier %? It's one of the more useful metrics for healers


EliphasNUIT:
Thanks!

PR #1142, opened 2026-04-23
Limit Hardened Crust on sulfuric geysers to normal mode only

Linkaaaaa:
From the feedback survey: 
> For Ura: in CM and LCM Sulfurics can't have their buff removed, yet the combat replay displays the buff over every Sulfuric with a very attention grabbing bouncing animation.


EliphasNUIT:
Thanks!

Issue #891, opened 2024-05-19
Combat Replay: Autoscroll the mechanics table as the replay is active


Linkaaaaa:
Add a checkbox to enable the mechanics table autoscrolling as the combat replay plays.

When autoscroll is selected, the table would:
- Automatically go to the correct line where the first mechanic timestamp is >= current combat replay time stamp.
- Scrolling the replay timeline left and right would move the table up and down.
- Selecting backwards on the replay would invert the scroll direction of the table.
- Clicking reset would take it back to the top.
- Remove the bootstrap tooltips once everything is functional.

The UI would just be this:

![7301](https://github.com/baaron4/GW2-Elite-Insights-Parser/assets/23245218/9d8144f5-0283-4d90-9cfa-9b8fc228e068)

Linkaaaaa:
Completed in #1145

PR #1145, opened 2026-04-27
Mechanics table automatic scrolling

Linkaaaaa:
Mechanics table auto scrolls while the replay is playing, can be disabled with the checkbox (off by default). #891
Fixed < ul> element nesting violations in 4 files.
Slight css updates to the fixed elements.


EliphasNUIT:
Thanks!

PR #1146, opened 2026-04-29
Added a new combat replay timeline

Linkaaaaa:
Current issues:
- Double clicking on a phase while the cursor is there doesn't take you to the start of the phase, you have to click away from the cursor for it to work.
- The main line doesn't function well with the highlighting on hover so it's not implemented right now.
- The colors for the light mode aren't set


EliphasNUIT:
Thanks!

PR #1149, opened 2026-05-02
Instance buff function and Lava Slam fix

Linkaaaaa:
Moved the instance buff logic to a function and fixed Lava Slam animation ending too early (had a + instead of -), the rest are white space automatic changes by visual studio...

### DIFF
--- a/GW2EIBuilders/Resources/htmlTemplates/Headers/tmplEncounter.html
+++ b/GW2EIBuilders/Resources/htmlTemplates/Headers/tmplEncounter.html
@@ -263,8 +263,13 @@
                 let description = instanceBuff.buff.description;
 
                 if (description) {
+                    let stacks = "<br>Max Stack(s) 1";
+                    if (description.includes(stacks)) {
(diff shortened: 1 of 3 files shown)


GW2EIBuilders/Resources/htmlTemplates/Headers/tmplEncounter.html
```diff
@@ -261,6 +258,19 @@ <h3 v-if="!isMultiEncounterLog" class="card-header text-center">
                 }
                 return;
             },
+            getBuffTooltip: function (instanceBuff) {
+                let title = instanceBuff.buff.name;
```
EliphasNUIT, line 262: What happened to not showing max stack if equal to 1?

Linkaaaaa: I've restored it because it was a bit jank, can put it back though

EliphasNUIT:
Thanks!

PR #1151, opened 2026-05-12
Added support to upload zevtc files to mistwarrior.com

virgilioneto:
Mist Warrior is a project focused on WvW and one of the features is the combined view of parsed logs using GW2_EI_log_combiner and a custom log dashboard in development by me.

This implementation works like the DPSReports uploader.


GW2EIMistWarrior/GW2EIMistWarrior.csproj
```diff
@@ -0,0 +1,36 @@
...
+        <DefineConstants>TRACING</DefineConstants>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='NoRewards|AnyCPU'">
+        <DefineConstants>DEBUG</DefineConstants>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <AnalysisMode>AllEnabledByDefault</AnalysisMode>
+        <Version>7.0.1</Version>
```
EliphasNUIT: Version 1, you can also put your alias in Authors

GW2EIParserCommons/ProgramHelper.cs
```diff
@@ -319,6 +321,13 @@ private string[] UploadOperation(FileInfo fInfo, ParsedEvtcLog originalLog, Oper
             originalController.UpdateProgressWithCancellationCheck("Wingman: Operation completed");
 
         }
+        if (Settings.UploadToMistWarrior)
+        {
```
EliphasNUIT, line 324: In GW2EIParserCommons\OperationControllers\OperationController.cs you'll find the following:
public bool DPSReportUploadTentative { get; internal set; }
public bool DPSReportUploadFailed { get; internal set; }

Its useage needs to be repeated for MistWarriorUpload so that the info can be passed through the CLI output.

GW2EIMistWarrior/MistWarriorUploadJsons/MistWarriorUploadObject.cs
```diff
@@ -0,0 +1,8 @@
+namespace GW2EIMistWarrior.MistWarriorUploadJsons;
+
+#pragma warning disable CS0649 // field never assigned to
+public class MistWarriorUploadObject
+{
+    public string? Id;
```
EliphasNUIT: I'm not familiar with MistWarrior APIs, any documentation? What does that ID mean? Something we would want to put in the HTML/JSON output?

virgilioneto: This ID means the returned identifier of the uploade log, but has no use. I'll remove and change the validation to be based on http status code only.

GW2EIParserCommons/ProgramHelper.cs
virgilioneto: Ok, I'll fix.

GW2EIMistWarrior/GW2EIMistWarrior.csproj
virgilioneto: Need to change the Company and Product too?

EliphasNUIT: No, that won't be necessary.

EliphasNUIT:
Looks good to me, thanks.

virgilioneto:
Thanks!

PR #1160, opened 2026-06-09
Added cave slublings ids and icons

Linkaaaaa:
Areki request, the slublings can enter the arena if left alive so it's better to display them


EliphasNUIT:
Thanks!

PR #1163, opened 2026-06-16
Added mechanic ticks to timeline

Linkaaaaa:
The ticks on the timeline represent the instances of the mechanic that happened for the selected player.
Button under the timeline that appears/disappears when a mechanic is selected/deselected.
Button above the mechanics table to deselect the mechanic.
Click the current select mechanic to deselect it.
Timeline text is updated to include MechanicName (Agent) when a mechanic is selected.

### DIFF
--- a/GW2EIBuilders/Resources/combatReplayTemplates/tmplCombatReplayAnimationControl.html
+++ b/GW2EIBuilders/Resources/combatReplayTemplates/tmplCombatReplayAnimationControl.html
@@ -202,7 +202,7 @@
                     </div>
                     <div v-if="!isTouchDevice && isMechanicSelected"
                          id="mechReset"
-                         @click="resetMechanic(selectedPhase);" class="btn btn-small"
+                         @click="resetMechanic();" class="btn btn-small"
                          :class="{'btn-dark': !light, 'btn-light': light}"
                          data-original-title="Remove mechanic ticks from timeline">
                         Reset Mechanic
@@ -530,7 +530,7 @@
                 }
                 return res;
             },
-            resetMechanic: function (phase) {
+            resetMechanic: function () {
                 $('#mechReset').tooltip('hide');
 
                 this.animationData.selectedMechanic = {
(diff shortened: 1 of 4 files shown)


GW2EIBuilders/Resources/combatReplayTemplates/tmplCombatReplayAnimationControl.html
```diff
@@ -483,7 +524,22 @@
...
+            resetMechanic: function (phase) {
+                $('#mechReset').tooltip('hide');
+
+                this.animationData.selectedMechanic = {
+                    actorId: null,
+                    actorName: null,
+                    name: null,
+                    times: [],
+                };
+                this.getFormattedPhaseText(phase);
```
EliphasNUIT: What's going on with phase here? A copy paste error?

Linkaaaaa: No i use it to remember which current phase is selected and only reset the mechanic text

EliphasNUIT: But that call does nothing, this.getFormattedPhaseText(phase) computes and returns a string, you don't do anything with it

Linkaaaaa: Right, i didn't notice the return was missing so it works without the need to call that method

EliphasNUIT: Yeah, that method will be automatically called when the view is refreshed when something changes.

GW2EIBuilders/Resources/combatReplayTemplates/tmplCombatReplayMechanicsList.html
```diff
@@ -209,6 +213,24 @@
                 this.followTimeline = false;
                 animator.stopAnimate(true);
                 animator.updateTime(mechanic.time);
+
```
EliphasNUIT, line 216: Should do an early return if isTouchDevice.

EliphasNUIT:
Thanks, I'll clean up the interactions.

PR #1167, opened 2026-06-27
Added guardian and revenant spear skills to autoattacks

Linkaaaaa:
The other specs seem to behave correctly


EliphasNUIT:
Thanks!

PR #1170, opened 2026-07-17
Adjusted Sulfuric Eruption radius

Linkaaaaa:
I've played around with the radius and it looks like 3400 matches with the Sulfuric Eruption Hit mechanic event, the decoration wave is roughly in the center of the players hitbox whenever it is selected in the table.


EliphasNUIT:
Thanks

PR #1177, opened 2026-08-23
Added missing -update command description help

Linkaaaaa:
We have never added a description for the -update command to the cli


EliphasNUIT:
Thanks.

PR #1178, opened 2026-08-27
Avalonia UI Migration

Linkaaaaa:
WinForm to AvaloniaUI migration

Added:
- [x] An X button to remove a log from the list
- [x] Dynamically update the parsing status as the traces arrive
- [x] A column to open log traces when enabled
- [x] Dynamic queued logs text
- [x] Click table to open file picker

Migrated:
- [x] All buttons on the main window
- [x] Drag and drop behavior
- [x] All of the settings
- [x] The updater window
- [x] Message windows for notifications
- [x] App icon

To review:
- [x] Discord upload
- [x] Settings tabs and categorizations


GW2EIParserCommons/ProgramHelper.cs
```diff
@@ -567,26 +567,29 @@ private static void CompressFile(string file, MemoryStream str, OperationControl
 
     private DirectoryInfo GetSaveDirectory(FileInfo fInfo)
     {
-        //save location
-        DirectoryInfo? saveDirectory;
-        if (Settings.SaveAtOut || Settings.OutLocation == null)
+        if (Settings.SaveAtOut)
```
EliphasNUIT: SaveAtOut is to save where the evtc file is , currently it throws when OutLocation is not defined when SaveAtOut is true, did you mean !Settings.SaveAtOut? 
Also previously if out location was not provided despite SaveAtOut being false, the app would default to evtc location, that is not the case anymore.

EliphasNUIT:
"Time (hours) textbox (what does this even do? I need to check)"
It is for populate from directory button, it restricts the fetched files to files younger than given time.

Linkaaaaa:
Can I move SupportedFileFormats from ParserHelpers to EICommons? Seems more fitting to me

EliphasNUIT:
But that is used inside EvtcParser.cs? Also GW2EIEvtcParser being the one with the knowledge of what file format it can parse/accept seems more logical to me.

EliphasNUIT:
"Dynamically update the parsing status as the traces arrive"
This is something I explicitly did not do in WinForms, we have a lot of traces and when multiple log parsing is enabled, it causes the UI to lag quite a bit as the main thread is put under pressure. I don't know, however, if Avalonia handles that better.

GW2EIParserCommons/ProgramHelper.cs
```diff
@@ -569,7 +569,7 @@ private DirectoryInfo GetSaveDirectory(FileInfo fInfo)
     {
         //save location
         DirectoryInfo? saveDirectory;
-        if (Settings.SaveAtOut || Settings.OutLocation == null)
+        if (Settings.SaveAtOut || Settings.OutLocation == null || string.IsNullOrWhiteSpace(Settings.OutLocation) || string.IsNullOrEmpty(Settings.OutLocation))
```
EliphasNUIT: Settings.OutLocation == null redundant when IsNullOrWhiteSpace and IsNullOrEmpty are used.

Linkaaaaa:
> But that is used inside EvtcParser.cs? Also GW2EIEvtcParser being the one with the knowledge of what file format it can parse/accept seems more logical to me.

Twice inside EvtcParser and 4 times inside Commons but I can leave it where it is, not a problem at all

> This is something I explicitly did not do in WinForms, we have a lot of traces and when multiple log parsing is enabled, it causes the UI to lag quite a bit as the main thread is put under pressure. I don't know, however, if Avalonia handles that better.

I'll try it out and if it lags I won't commit to having it

Linkaaaaa:
Don't need to check the code as the traces aren't fully migrated but what is there *mostly* works.

I do have a few questions

1: Can I move the checkboxes for Traces and Auto Discord Batch and the textbox for the time limit into the settings window? They are settings at the end and they don't belong to the main window in my opinion, additionally they entangle the view models unnecessarily if they remain in the main window. The Send all to Discord button would remain as that is an active action.
If i can move them, any preferences of their location?

2: Would it make sense to switch all of the settings to be using GW2EIParserCommons.Properties.Settings.Default in a transition like this? Or does it make it more annoying for you to review settings files?

EliphasNUIT:
1 - Traces and Auto why not. Time directly impacts the Populate From Directory button, it does not make sense to separate them. Traces can go to General, Auto to Upload. As long as we are talking about the avalonia project obviously, please don't touch GW2EIParser and GW2EIParserCLI in this PR if not due to a renaming or something like that ^^
2 - I'm not sure what you mean by that. Settings are all declared in Commons/Settings/Settings.cs already.

Linkaaaaa:
1: Yes I don't want to touch anything regarding the old ones unless strictly necessary (like adding a list for the traces files in the latest commit)
2: I see Settings.cs but we have a bunch of them inside Commons.ProgramSettings and others that aren't there like the application traces, auto Discord batch and hour limit

EliphasNUIT:
2 - Those 3 are specific to the UI application, so they shouldn't be in common, they are not relevant to that project. ProgramSettings is something for external useage, so that someone can include Commons (and dependencies) into their own project and easily configure ProgramHelper, Properties.Settings.Default is a little bit akward to use.

Linkaaaaa:
Looks like the UI updates just fine, it doesn't seem laggy to me (although I can't parse many more logs than this at the same time, the pc can't handle it)

https://github.com/user-attachments/assets/bcb769df-2631-4ded-939e-4fe02fb95530

EliphasNUIT:
I'll try on my end later, I should be able to parse 32 at the same time.

GW2EIParserAvalonia/ViewModels/SettingsViewModel.cs
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    public void ApplyLoadedSettings(string path)
+    {
+        CustomSettingsManager.ReadConfig(path);
+
+        var loadedSettings = CustomSettingsManager.GetProgramSettings();
+
+        _settings.CopyFrom(loadedSettings);
```
EliphasNUIT: This feel a little convoluted, I'll improve GetProgramSettings to make things easier and also simplify existing code.

Linkaaaaa: Feel free to commit to my branch, there is SO MUCH clean up to do anyway lol I should have done the migrating manually. I don't have anything else that hasn't been pushed so there won't be collisions

EliphasNUIT: I also want to simplify in existing code, that "ApplySettings" method is somewhat useless

Linkaaaaa: I've merged with master and won't touch anything else for tonight so you can try it out

Linkaaaaa:
Small note: I've renamed "Clear Failed" button to "Clear Uncompleted", for a while I kept thinking it was supposed to remove logs that had a failure result, instead it clears logs that failed to parse and this is clearer to me

Linkaaaaa:
I believe most of the functionalities should be done now, there are still application traces to be added in various places but it should function.

Linkaaaaa:
I've tested building the project for Linux through a friend that had it and it runs natively without any wine/proton installed so it's working correctly as intended.

I have not tested the Discord functionalities as I don't have a webhook setup to try it out. 

Was thinking of changing how the settings are laid out (merging HTML/CSV/JSON tabs for example) but for now I'll leave them as they are so you know what is where for testing.

After this project is completed, I'm thinking of adding an inspect button to the table that will be wired with the evtcparser to display all of the combat items. That will be a way bigger update and it's out of scope of the current migration and I don't have an UI idea for it yet.

Let me know when you will have time to go through all of this.

GW2EIParserAvalonia/ViewModels/SettingsViewModel.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+            using StreamWriter sw = File.AppendText(_applicationTraceFileName);
+            sw.WriteLine(message);
+        }
+    }
+
+    public void ApplyLoadedSettings(string path)
+    {
+        CustomSettingsManager.ReadConfig(path);
+        var loadedSettings = CustomSettingsManager.GetProgramSettings();
```
EliphasNUIT: you can give _settings to GetProgramSettings as an argument and remove LoadFromProgramSettings

GW2EIParserAvalonia/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace GW2EIParserAvalonia;
+
+sealed class Program
+{
+    // Initialization code. Don't use any Avalonia, third-party APIs or any
+    // SynchronizationContext-reliant code before AppMain is called: things aren't initialized
+    // yet and stuff might break.
+    [STAThread]
+    public static void Main(string[] args) => BuildAvaloniaApp()
```
EliphasNUIT: Missing quite a bit of stuff compared to Program.cs from GW2EIParser (update path temp cleanup, argument handling for conf and evtc files, help, etc).

GW2EIParserAvalonia/ViewModels/MainWindowViewModel.cs
```diff
@@ -0,0 +1,497 @@
...
+            else
+            {
+                UpdateLogState(logFile);
+            }
+
+            _parserService.GenerateTraceFile(logFile.Operation);
+
+            if (logFile.State != OperationState.Complete)
+            {
+                logFile.Operation.Reset();
```
EliphasNUIT: ResetContent()

GW2EIParserAvalonia/ViewModels/MainWindowViewModel.cs
```diff
@@ -0,0 +1,497 @@
...
+            if (logFile.State != OperationState.ClearOnCancel)
+            {
+                logFile.Operation.ToUnCompleteState(reason);
+            }
+        }
+        finally
+        {
+            _runningCount--;
+
+            if (logFile.State == OperationState.ClearOnCancel)
```
EliphasNUIT, line 333: Something should dispose the cancellationToken properly, instead of it being done when RunState or ResetContent is called. Currently, on successful logs parsed only once, the token is never disposed.

EliphasNUIT: Maybe create the token in this method and pass it to Operation via ToRunState? And dispose it in finally. Then you could remove Token lifecycle handling from Operation. Not a big fan of Operation handling task related objects' lifecycle

EliphasNUIT:
Put down some remarks regarding what was immediately apparent to me. I'll play with this a little on my side, need to familiarise myself with Avalonia APIs and execution flows before I can eventually make more technical remarks.

Linkaaaaa:
Let me try to address those points before continuing to review, the command line stuff in of itself is making this quite extra complex to setup

GW2EIParserAvalonia/Program.cs
Linkaaaaa: I had missed all of that and now it has been added back, running `GW2EIParserAvalonia.exe "path\file.zevtc"` correctly opens the UI and adds the file to the table. I think all of this should live elsewhere since a chunk of this is already in the CLI project though, probably something that should be looked at later on.

GW2EIParserAvalonia/ViewModels/MainWindowViewModel.cs
Linkaaaaa: I've changed the way the token is disposed and now it should work more correctly

Linkaaaaa:
You can try playing with it now and feel free to commit directly any change that you see fit, I'll not touch anything until you send reviews or similar

GW2EIParserAvalonia/Services/SettingsService.cs
```diff
@@ -0,0 +1,50 @@
...
+    private static readonly JsonSerializerOptions JsonOptions = new()
+    {
+        WriteIndented = true
+    };
+
+    public SettingsService()
+    {
+        var directory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "GW2EliteInsights");
+        Directory.CreateDirectory(directory);
+        _filePath = Path.Combine(directory, "settings.json");
```
EliphasNUIT: Any reason why a brand new setting file format was introduced? If settings.json exists, it overrides .conf file given by the user.

GW2EIParserAvalonia/Views/SettingsWindow.axaml.cs
```diff
@@ -0,0 +1,145 @@
...
+        Close();
+    }
+
+    private void CancelButton_Click(object? sender, RoutedEventArgs e)
+    {
+        _trace.Add("UI: Settings cancelled");
+        Close();
+    }
+
+    private async void LoadButton_Click(object? sender, RoutedEventArgs e)
```
EliphasNUIT: Function to dump current settings into a .conf file is missing

Linkaaaaa:
One small thing, I'd appreciate if we would not change the way the old ui works anymore since this is aiming to replace it anyway. Having to check your changes, merge them, understand what got changed and update the new code to match the behavior is just extra work for both of us since you also will have to read my changes and control that they are on pair with yours.

EliphasNUIT:
Don't think anything changed on the old UI apart from some renaming? And a bug fix where passing from state Queue to Cancel/ClearOnCancel was not working but it is super hard to reproduce, tasks are rarely delayed enough for a human to be able to click cancel while just on Queue.

I've created test-pr branch on my side where I play with it and change some stuff.

Linkaaaaa:
I'm talking about https://github.com/baaron4/GW2-Elite-Insights-Parser/commit/cabe2522d4fe376b7f8992a6530ddde930b79917 adding two new OperationState, removal of IsBusy, changing RunOperation etc. I'm trying to adapt all of these changes to the new code now but then you will also have to review them back, so it's just extra work for both

EliphasNUIT:
I'll already brought those to my test branch, I'll merge it to the PR when I'm satisfied with it.

Linkaaaaa:
Ok then I won't continue on my end in this case

GW2EIParserAvalonia/Services/SettingsService.cs
Linkaaaaa: No that's just wrong, needs to be fixed

EliphasNUIT: Yeah, I completely removed all that json shennanigans ^^

Linkaaaaa:
Now that the ui seems to handle live updates really fast and smooth, we could have a debug window that shows application and log traces as they come in instead of having to open the files each time, but I don't have a plan of where this would be or how it would look like right now.
It's not something necessary for this current PR anyway, just noting down this idea.

EliphasNUIT:
Hmm they are mainly for debug purposes so I don't think there is a major added values for the average user (particularly for a feature that updates dynamically as file is being processed), we could have it generated dynamically though, the trace button opens a window with the traces in it with a button to dump it into a file for sharing.

Linkaaaaa:
Yeah the window would be there only in debug, could also have an option in the settings to show debug features, just ideas thrown out there to think about

EliphasNUIT:
Webhooks are working as expected.

EliphasNUIT:
made settings window non blocking + save settings button + external script folder selection button. There shouldn't be any missing features, I'll merge this soon I believe.

Linkaaaaa:
Do we want to merge HTML + CSV + JSON tabs under "Formats" or "Output"? Having the 3 separated seems unnecessary
If you want I can make this change

EliphasNUIT:
Sure

Linkaaaaa:
Two notes from my friend with Linux on hand (I still don't have a setup for it)

When building with `dotnet publish -c Release -r linux-x64 --self-contained true` gives `error NETSDK1152: Found multiple publish output files with the same relative path` which he fixed by adding `<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles>` inside `<PropertyGroup>` in `GW2EIParserAvalonia.csproj`

Second, it gave an error in CommandWindow:

`Unhandled exception. System.DllNotFoundException: Unable to load shared library 'kernel32.dll' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: ` 

which he fixed by wrapping this section inside `if (OperatingSystem.IsWindows())`

```c#
if (IsRedirected(GetStdHandle(StandardHandle.Output)))
{
    _ = Console.Out;
}

if (!AttachConsole(AttachParentProcess))
{
    AllocConsole();
}

AttachConsole(AttachParentProcess);
```

I don't know if these are the appropriate fixes though

Linkaaaaa:
Also maybe it's worth setting up some github actions that build for the various platforms now? Not for this PR anyway and I haven't used them before

EliphasNUIT:
The 2nd point is correct yes. For the first point, no, we need to solve the source of the problem, we won't disable that, what are the offending files/paths?

Linkaaaaa:
GuildWars2EliteInsightsUpdater.deps.json
GuildWars2EliteInsightsUpdater.runtimeconfig.json

And you also get NETSDK1082 No runtime pack for Microsoft.WindowsDesktop.App.Winforms for linux-x64, but this is to be expected

Also you get all of these by running the command on windows as well so you should be able to reproduce them

Linkaaaaa:
Current small bug: Cancel All button never activates (not when clicking Parse, Re-Parse or Parse All)

EliphasNUIT:
Can't  reproduce with Parse All, that's working nicely.

Linkaaaaa:
I first parsed the log with the Parse button and then Parse All wasn't activating Cancel All

EliphasNUIT:
in QueueOrRunOperation, UpdateButtonStates is called at the start, should be at the end

PR #1180, opened 2026-08-30
Arkk improvements

Zerthox:
This adds the reactor gadgets (only logged since halfway through 2024) to the combat replay and adjusts the map to be more in line with their actual positions.

The electrocuted areas might also be worth adding. So far I did not see any events related to them, their visual might simply be part of the reactor models.

### DIFF
--- a/GW2EIEvtcParser/ParserHelpers/IDs/SpeciesIDs.cs
+++ b/GW2EIEvtcParser/ParserHelpers/IDs/SpeciesIDs.cs
@@ -641,6 +641,7 @@ public static class SpeciesIDs
         DOC = 16657,
         CHOP = 16552,
         ProjectionArkk = 17613,
+        ReactorArkk = 17823, // gadget
         // - Ai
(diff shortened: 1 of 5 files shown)


GW2EIEvtcParser/ParserHelpers/IDs/SpeciesIDs.cs
```diff
@@ -641,6 +641,7 @@ public enum TargetID : int
         DOC = 16657,
         CHOP = 16552,
         ProjectionArkk = 17613,
+        ReactorArkk = 17823, // gadget
```
EliphasNUIT: Gadget ids should be stabilized, I see they are involved with a special buff, should be pretty straightforward

Zerthox: I'll correct my comment, they are character agents not gadgets.

EliphasNUIT:
Thanks!

PR #1184, opened 2026-09-06
Application traces fix

Linkaaaaa:
When parsing multiple logs I was getting an I/O Exception because the threads were attempting to write to the same file at the same time.
I've also added a timestamp to each message so we can see the timings of operations.

How the log looks like:


EliphasNUIT:
Thanks.

PR #1186, opened 2026-09-06
EVTC Inspector

Linkaaaaa:
Initial draft so you can take a look.

I'm not sure how I want the UI to be yet, I'll probably copy the setup from the other inspector or something like that


GW2EI.Library/GW2EI.Services/GW2EIEvtcParser/ParsedEvtcLog.cs
```diff
@@ -13,6 +13,7 @@ public class ParsedEvtcLog
     public readonly LogData LogData;
     public readonly AgentData AgentData;
     public readonly SkillData SkillData;
+    public readonly IReadOnlyList<CombatItem> CombatItems;
```
EliphasNUIT: We are not going to conserve duplicate combat items on the basis that we maybe will show them, the inspector should show EI's combat items, it is ok to keep it while testing/active dev, but this can't stay in the final product.

GW2EI.Library/GW2EI.Services/GW2EIEvtcParser/ParsedData/Skills/SkillItem.cs
```diff
@@ -49,6 +49,7 @@ public bool IsAutoAttack(ParsedEvtcLog log) => AA
     private readonly WeaponDescriptor? _weaponDescriptor;
     public bool IsWeaponSkill => _weaponDescriptor != null;
     internal readonly GW2APISkill? ApiSkill;
+    public GW2APISkill? ApiSkillInfo => ApiSkill;
```
EliphasNUIT: Can simply make ApiSkill a public readonly attribute

GW2EI.Executables/GW2EI.Applications/GW2EIParserAvalonia/Services/ParserService.cs
```diff
@@ -11,6 +12,7 @@ public sealed class ParserService : IDisposable
 {
     private readonly ProgramHelper _programHelper;
     public ProgramSettings Settings { get; }
+    public ParsedEvtcLog? ParsedLog => _programHelper.InspectLog;
```
EliphasNUIT: Should be stored on InspectorOperationController so that we don't have to reparse everytime inspect is called, we can re-use already parsed evtc log if present on LogFileViewModel.

GW2EI.Library/GW2EI.Services/GW2EIParserCommons/ProgramHelper.cs
```diff
@@ -474,6 +475,48 @@ private string[] UploadOperation(FileInfo fInfo, ParsedEvtcLog originalLog, Oper
...
+                ComputeBuff = false,
+                ComputeDamage = false,
+                ComputeCast = false,
+                ComputeMechanics = false,
+                DetailedWvWParse = false,
+            },
+                                        APIController);
+
+            //Process evtc here
+            InspectLog = parser.ParseLog(operation, fInfo, out var failureReason, !Settings.SingleThreaded && HasFormat());
```
EliphasNUIT: false instead of !Settings...

EliphasNUIT: and failure reason should be handled

GW2EI.Library/GW2EI.Services/GW2EIParserCommons/ProgramHelper.cs
```diff
@@ -474,6 +475,48 @@ private string[] UploadOperation(FileInfo fInfo, ParsedEvtcLog originalLog, Oper
         return uploadresult;
     }
     #endregion UPLOAD
+    public void Inspect(OperationController operation)
```
EliphasNUIT: Should return ParsedEvtcLog? instead of void, no reason to store it on a helper class

Linkaaaaa:
This is already becoming very big and it's like 5% of what it should be as end result, we will probably have to split it up into a separated project

EliphasNUIT:
Thanks!

PR #1189, opened 2026-09-11
Content GUID tab changes

Linkaaaaa:
- All GUIDs are now sorted by ContentID
- Reordered the tabs so that the most used are on the first ones (effects, markers)
- Added filtering by ContentID and ContentGUID shared across all tabs

(most of the rows changes are indentation and reordering)

### DIFF
--- a/GW2EI.Executables/GW2EI.Applications/GW2EIParserAvalonia/Inspector/ViewModels/InspectorViewModel.cs
+++ b/GW2EI.Executables/GW2EI.Applications/GW2EIParserAvalonia/Inspector/ViewModels/InspectorViewModel.cs
@@ -17,12 +17,20 @@ public partial class InspectorViewModel : ObservableObject
     private AgentDataModel? selectedAgent;
     [ObservableProperty]
     private SkillDataModel? selectedSkill;
+
+    // Events tab filters
     [ObservableProperty]
     private string? skillIdFilter;
     [ObservableProperty]
     private string? skillNameFilter;
     [ObservableProperty]
     private string? guidFilter;
+
+    // Content GUID tab filters
+    [ObservableProperty]
+    private string? contentIdFilter;
+    [ObservableProperty]
+    private string? contentGuidFilter;
 
     private readonly IReadOnlyList<TimeCombatEvent> _allTimeEvents;
     private readonly IReadOnlyList<NonTimeCombatEvent> _allNonTimeEvents;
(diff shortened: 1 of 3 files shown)


Linkaaaaa:
I've added a parse fix for whisper of jormag, after the merge with master the jump mechanic conflicted with the new one I've added

GW2EI.Executables/GW2EI.Applications/GW2EIParserAvalonia/Inspector/ViewModels/InspectorViewModel.cs
```diff
@@ -69,13 +85,13 @@ public InspectorViewModel(RawEvtcLog log)
...
 
         var contentGUIDEvents = _allNonTimeEvents.OfType<IDToGUIDEvent>().Where(x => x.IsValid).ToList();
-        Skills = contentGUIDEvents.OfType<SkillGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
-        Effects = contentGUIDEvents.OfType<EffectGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
-        Markers = contentGUIDEvents.OfType<MarkerGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
-        Species = contentGUIDEvents.OfType<SpeciesGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
-        Teams = contentGUIDEvents.OfType<TeamGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
-        Emotes = contentGUIDEvents.OfType<EmoteGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
-        Transformations = contentGUIDEvents.OfType<TransformationGUIDEvent>().Select(x => new ContentGUIDModel(x)).ToList();
+        Skills = contentGUIDEvents.OfType<SkillGUIDEvent>().Select(x => new ContentGUIDModel(x)).OrderBy(skill => skill.ContentID).ToList();
```
EliphasNUIT, line 88: It would make more sense for the default sort to be based on a stable property. I have implemented IComparable on GUID so you can to an OrderBy(x => x.GUID)

Linkaaaaa: I know that the ID is not stable but while working on the replay I found it easier to check an effect event and then sort by ID and just remember the 4-5 numbers than trying to find the GUID itself, although this also adds the textbox filters, so I'm not too sure what is best for usability

EliphasNUIT:
Thanks!

PR #1190, opened 2026-09-14
Update guid filter

Linkaaaaa:
Removed .ToString() from EventModel
Changed filter to use .ToString() only if the GUID failed to parse (< 32 or wrong characters)
Updated input boxes to only accept 32 characters
Changed ValidateHex to use Guid.TryParseExact
Compacted and ordered operators
Added Content ID filter to Events tab
Moved Skill ID on the same line as Skill Name


EliphasNUIT:
Thanks!

PR #1191, opened 2026-09-16
Added new relics

Linkaaaaa:
Missing icons (not on the wiki yet and I don't know their api ids)
Relic of the last Tyrant cast proc is tricky but the current check works, I didn't see another way of adding it that is more reliable

Files
[voe relics set 4.zip](https://github.com/user-attachments/files/32277938/voe.relics.set.4.zip)


GW2EI.Library/GW2EI.Services/GW2EIEvtcParser/EIData/DamageModifiers/CommonDamageModifiers/GearDamageModifiers.cs
```diff
@@ -110,5 +112,6 @@ internal static class GearDamageModifiers
             .WithBuilds(GW2Builds.May2026NewRelicsAndOtter),
         new BuffOnActorDamageModifier(Mod_RelicOfTheDoyen, Resolution, "Relic of the Doyen", "-7% (not stacking with similar damage mods)", DamageSource.Incoming, -7.0, DamageType.Strike, DamageType.All, Source.Gear, ByPresence, ItemImages.RelicOfTheDoyen, DamageModifierMode.sPvP)
             .WithBuilds(GW2Builds.May2026NewRelicsAndOtter),
+        new BuffOnActorDamageModifier(Mod_RelicOfTheCurator, Protection, "Relic of the Curator", "-20% (not stacking with similar damage mods)", DamageSource.Incoming, -7.0, DamageType.Strike, DamageType.All, Source.Gear, ByPresence, ItemImages.RelicOfTheCurator, DamageModifierMode.All)
```
EliphasNUIT: This is effectiveness increase no? So it'd -53 or -39.6 depending on how the increase is applied. The description should also indicate that this would replace the protection damage mod.

Linkaaaaa: I've just copied the one above atm, don't know how the relic works and we don't have the skills dump (ask rennorb for it?)

GW2EI.Library/GW2EI.Services/GW2EIEvtcParser/EIData/ProfHelpers/ProfHelper.cs
```diff
@@ -115,6 +115,20 @@ internal static class ProfHelper
         #region Relics
         new DamageCastFinder(RelicOfShacklesDamageSkill, RelicOfShacklesDamageSkill)
             .UsingOrigin(InstantCastFinder.InstantCastOrigin.Gear),
+        new DamageCastFinder(RelicOfTheLastTyrantDamage, RelicOfTheLastTyrantDamage)
+            .UsingChecker((hde, combatData, agentData, skillData) => {
+                // When the effect procs, it is on the target of the damage
+                // The Src of the effect is the target
+                // The Dst of the effect is unknown
+                if (combatData.TryGetEffectEventsByGUID(EffectGUIDs.RelicOfTheLastTyrant, out var effects))
+                {
```
EliphasNUIT: I feel like this has the potential to generate false positives as the effect has no connection to the actual src, perhaps a damage cast finder associated with a BuffRemoveAll event for TyrantsFuryBuff on the damage source would be safer?

EliphasNUIT:
Thanks.

PR #1195, opened 2026-09-19
Nexus of Eternity arena image

Linkaaaaa:
From @LunusLupus 

Most line changes are indentation, only changed a couple of colors and added the Worldpiercer projectile


EliphasNUIT:
Thanks!

PR #1196, opened 2026-09-19
Updated readme

dii2r2:
Just a few documentation changes, I used a dhuum log just like the old screenshots

- Added missing cli field in the json object
- Removed XML settings
- Updated all Elite Insights images
- Added missing descriptions for the new tabs that were added since last update
- Combined Players and Targets together since its a single component now and updated the description

If I missed anything or got anything wrong let me know, I'll try to fix it


Linkaaaaa:
Could change the UI and Settings images with the new Avalonia app, could also add a screenshot of the new inspector and document that

EliphasNUIT:
Thanks!

PR #1198, opened 2026-09-20
Update readme Part 2

dii2r2:
Apologies @EliphasNUIT I should've commented that I was adding the suggested changes from Linka plus a few more

The changes for part 2 are:
- documented the Evtc Inspector as suggested by @Linkaaaaa but I'm not entirely sure about their descriptions since I don't use the inspector as much as you guys or if you want even all 5 images of each of the tabs, i could use some input here or feel free to edit it the descriptions
- updated light and dark themes and the console usage images
- Tested Avalonia on a fresh Windows install, Debian, Arch and MacOS and updated the setup instructions for every OS they are now separated
- Also separated the CLI from the UI since its a lot simpler and to avoid the common questions
- I tested the UI from console as it was suggested before but it doesn't seem to work for Avalonia so I removed it I can add it back in and mark it that it only works for WinForms
- Adjusted the image sizes to be 60% width and in the cases where there were 2 stacked on top of each other i used a div with style flex to just put them side by side, in UI Usage and Players & Targets sections
- Added a requirement section since it won't prompt users on Linux/MacOS 
- Added the fancy notes and important blocks to draw a bit more attention to it
- Added extra images for the other settings tabs these can maybe be removed i didn't add them in the readme as i thought it was too much

If you want i can also add where to find the arcdps common paths on Linux there is usually 2 different ones by default one for steam and the other is from third party game launchers like UMU or Lutris

If you want any changes or if you spot any mistakes let me know


EliphasNUIT:
No problem, that request felt out of scope of that initial PR anyway, this is not just cleanup territory.

README.md
```diff
@@ -410,6 +448,43 @@ Download and install this extenstion to gain access to healing statistics. Read
...
+The EVTC Inspector allows you to inspect metadata and events that occurred during the encounter.
+
+Selecting any entry from the tables in Events, Agents Data or Skills Data will display the detailed information about it. 
+
+The table columns can be sorted and resized and selecting a row and using `Ctrl + c` will copy all of the columns for the row.
+
+### Events
+<img src="./docs/Images/inspectorEvents.png" width="60%">
+
+Contains every event that occured during the encounter, can be filtered by a specific `Skill ID`, `Skill Name`, `GUID`, `Agent` and you can select which event types you would like to include/exclude from the table with the event tree.
```
Linkaaaaa: `[...] can be filtered by a specific Skill, Content GUID or Agent [...]`
`[...] event tree. You can click an event in the list to inspect its content.`

On master and with my current PR there are already more filters than the current image (GUID ID, Agent By)

README.md
```diff
@@ -25,47 +25,88 @@ We suggest following [this guide](https://snowcrows.com/guides/arcdps/arcdps) wr
...
-1. Drag and drop one or multiple .evtc, .evtc.zip, or .zevtc files onto the program.
+1. Use either `Add Files` or `Populate from directory` buttons or drag and drop anywhere on the window one or multiple .evtc, .evtc.zip, or .zevtc files onto the program.
 
-2. Click parse.
+2. Click `Parse`.
 
-3. Click open when parsing done.
+3. Click `Open` when parsing done.
 
 You can change the settings at any time using the Settings window.
```
Linkaaaaa: We can change this line to something like
`You can use the Settings window to modify the parser output. The settings changes will not be applied to logs already parsing.`

dii2r2:
I forgot to mention i also have the AUR packages ready for arch linux so they would just be able to install it via `yay` or `paru` and dotnet is handled automatically as well.

We can mention it in the readme that there is a package available for arch in the AUR but completely up to you as its a specific distro but no maintenance from you would be required, I will be maintaining them either way as i use it personally and if they look they can find it easily too

Linkaaaaa:
Thank you for doing these updates btw, I was planning to do them in November when the WinForm UI will be deprecated/removed.

README.md
dii2r2: I just checked it out I'll wait until its merged to update the images and add the rest of the filters, idk if you're done or adding anything else to that pr

in the meantime I applied the changes you suggested

Linkaaaaa: Not adding anything to that one, tomorrow I want to make the events counter update with the filters so it shows something like this
BuffApplyEvent 45390 (2349) or BuffApplyEvent (2349) 45390
It's not something that will affect your screenshot anyway

Linkaaaaa:
Are you able to edit https://github.com/baaron4/GW2-Elite-Insights-Parser/wiki ? Most of the content of the readme should also be updated on the wiki, that hasn't been touched since 8 years lol

dii2r2:
> Are you able to edit https://github.com/baaron4/GW2-Elite-Insights-Parser/wiki ? Most of the content of the readme should also be updated on the wiki, that hasn't been touched since 8 years lol

I just realised there is a wiki lol, and yes it seems anyone with a github account can edit it since its not restricted to collaborators

EliphasNUIT:
Thanks!

PR #1200, opened 2026-09-21
updated release linux workflows

dii2r2:
I tested the program locally it all worked correctly, they can be installed simply by double clicking

Most notable change is the setting TargetLatestRuntimePatch to false since some distro's wont have the very last version available of dotnet, i also set the [rollFoward](https://learn.microsoft.com/en-us/dotnet/core/versions/selection) to latestPatch i think its correct but please confirm

.NET8 installation is handled on both deb and rpm now and when we move to .NET10 we just need to update the version


dii2r2:
forgot to add i will update the readme tomorrow to mention the deb/rpm and arch pkgs if you'd like

EliphasNUIT:
Thanks!

